PR #4, opened 2019-06-05
Implement swift-metrics

### LINKED ISSUE #3: Implement swift-metrics
### Feature Request

We should implement the workflow laid out by [`swift-metrics`](https://github.com/apple/swift-metrics) which will allow this to be plugged in as one of the backends services can choose to utilize.

Support the SSWG standardization.

#### Feature Description

This will require a bump in major version since it will drastically change the API.

MrLotU:
Fixes #3 😄


ktoso:
Awesome work here in general — glad we managed to polish the final bits together during try swift :) I’ll plan to do a full review here during the week 👍

Sources/Prometheus/MetricTypes/Gauge.swift
```diff
@@ -1,9 +1,11 @@
...
+import NIOConcurrencyHelpers
+
+/// Prometheus Gauge metric
 ///
 /// See https://prometheus.io/docs/concepts/metric_types/#gauge
-public class Gauge<NumType: Numeric, Labels: MetricLabels>: Metric, PrometheusHandled {
+public class PromGauge<NumType: Numeric, Labels: MetricLabels>: Metric, PrometheusHandled {
     /// Prometheus instance that created this Gauge
-    internal let prometheus: PrometheusClient
+    internal weak var prometheus: PrometheusClient?
```
ktoso: Not super sure they really have to be `weak` to be honest... are we sure about that?
In an app lifecycle you'd generally boostrap the metrics into the `MetricsSystem` and then it's there "forever"; what is the weak gaining us?

Sources/Prometheus/MetricTypes/Summary.swift
```diff
@@ -1,8 +1,11 @@
+import NIOConcurrencyHelpers
+
 /// Default quantiles used by Summaries
 public var defaultQuantiles = [0.01, 0.05, 0.5, 0.9, 0.95, 0.99, 0.999]
 
 /// Label type Summaries can use
 public protocol SummaryLabels: MetricLabels {
+    /// Quantile
     var quantile: String { get set }
```
ktoso: Some other comment maybe? would this be "human readable description of the quantile" perhaps?

Sources/Prometheus/MetricTypes/Summary.swift
```diff
@@ -1,8 +1,11 @@
+import NIOConcurrencyHelpers
+
 /// Default quantiles used by Summaries
 public var defaultQuantiles = [0.01, 0.05, 0.5, 0.9, 0.95, 0.99, 0.999]
```
ktoso: terrifying public global var! Please make it a `let` or otherwise we'd need to protect potential concurrent modifications. (or move it into a class and protect it there)

Sources/Prometheus/Prometheus.swift
```diff
@@ -89,13 +112,17 @@ public class PrometheusClient {
...
         initialValue: T = 0,
-        withLabelType labelType: U.Type) -> Gauge<T, U>
+        withLabelType labelType: U.Type) -> PromGauge<T, U>
     {
-        let gauge = Gauge<T, U>(name, helpText, initialValue, self)
-        prometheusQueue.async(flags: .barrier) {
+        return self.lock.withLock {
+            if let type = metricTypeMap[name] {
+                precondition(type == .gauge, "Label \(name) was associated with \(type) before. Can not be used for a gauge now.")
+            }
```
ktoso: Very good 👍

Sources/Prometheus/Prometheus.swift
```diff
@@ -199,30 +234,17 @@ public class PrometheusClient {
...
-        return info
-    }
 }
+
+/// Prometheus specific errors
+public enum PrometheusError: Error {
+    /// Thrown when a user tries to retrive
+    /// a `PromtheusClient` from `MetricsSystem`
+    /// but there was no `PrometheusClient` bootstrapped
+    case PrometheusFactoryNotBootstrapped
```
ktoso: AFAIK style is to have those lower case, or am I wrong (I totally could be wrong here)?

Sources/Prometheus/Utils.swift
```diff
@@ -1,28 +1,29 @@
 import Foundation
 
-/// Prometheus global dispatch queue
-internal let prometheusQueue = DispatchQueue(label: "prometheus.internal", attributes: .concurrent)
-
```
ktoso: glad that we don't need to use dispatch for the lib <3

Sources/PrometheusExample/main.swift
```diff
@@ -70,34 +111,5 @@ for _ in 0...Int.random(in: 100...1000) {
...
-    group.leave()
-}
-
-group.notify(queue: .main) {
-    exit(EXIT_SUCCESS)
-}
-
-dispatchMain()
+let metrics = try! MetricsSystem.prometheus().getMetrics()
+print(metrics)
```
ktoso: nice example app 👍

Sources/PrometheusMetrics/Exports.swift
```diff
@@ -0,0 +1,2 @@
+@_exported import Prometheus
+@_exported import CoreMetrics
```
ktoso: Should we be doing these @weissi  or better avoid?

Sources/PrometheusMetrics/PrometheusMetrics.swift
```diff
@@ -0,0 +1,307 @@
...
+
+public extension MetricsSystem {
+    /// Get the bootstrapped `MetricsSystem` as `PrometheusClient`
+    ///
+    /// - Returns: `PrometheusClient` used to bootstrap `MetricsSystem`
+    /// - Throws: `PrometheusError.PrometheusFactoryNotBootstrapped`
+    ///             if no `PrometheusClient` was used to bootstrap `MetricsSystem`
+    static func prometheus() throws -> PrometheusClient {
+        guard let prom = self.factory as? PrometheusClient else {
+            throw PrometheusError.PrometheusFactoryNotBootstrapped
```
ktoso: Could we carry information like `throw PrometheusError.PrometheusFactoryNotBootstrapped(boostrappedWith: "\(self.factory)")` so users know which library is messing with their bootstrap?

Sources/PrometheusMetrics/PrometheusMetrics.swift
```diff
@@ -0,0 +1,307 @@
...
+        }
+        return prom
+    }
+}
+
+// MARK: - Labels
+
+/// A generic `String` based `CodingKey` implementation.
+private struct StringCodingKey: CodingKey {
+    /// `CodingKey` conformance.
```
ktoso: is it typical to comment on all conforming things?  I think I'd drop those comments hm...
Although I could be missing a Swifty pattern about this?

Tests/SwiftPrometheusTests/PrometheusMetricsTests.swift
```diff
@@ -0,0 +1,137 @@
...
+        my_histogram_bucket{myValue="labels", le="0.75"} 0.0
+        my_histogram_bucket{myValue="labels", le="1.0"} 0.0
+        my_histogram_bucket{myValue="labels", le="2.5"} 0.0
+        my_histogram_bucket{myValue="labels", le="5.0"} 1.0
+        my_histogram_bucket{myValue="labels", le="7.5"} 1.0
+        my_histogram_bucket{myValue="labels", le="10.0"} 1.0
+        my_histogram_bucket{myValue="labels", le="+Inf"} 1.0
+        my_histogram_count{myValue="labels"} 1.0
+        my_histogram_sum{myValue="labels"} 3.0
+        """)
```
ktoso: ❤️

Sources/Prometheus/MetricTypes/Counter.swift
```diff
@@ -88,10 +96,12 @@ public class Counter<NumType: Numeric, Labels: MetricLabels>: Metric, Prometheus
...
-            return self.metrics[labels] ?? initialValue
-        } else {
-            return self.value
+        return self.lock.withLock {
+            if let labels = labels {
+                return self.metrics[labels] ?? initialValue
+            } else {
+                return self.value
+            }
         }
```
ktoso: 👍

Sources/Prometheus/MetricTypes/Histogram.swift
```diff
@@ -1,8 +1,11 @@
+import NIOConcurrencyHelpers
+
 /// Default buckets used by Histograms
 public var defaultBuckets = [0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1.0, 2.5, 5.0, 7.5, 10.0, Double.greatestFiniteMagnitude]
 
 /// Label type Histograms can use
 public protocol HistogramLabels: MetricLabels {
+    /// Bucket
     var le: String { get set }
```
ktoso: is this `le` on purpose?

Sources/Prometheus/MetricTypes/Metric.swift
```diff
@@ -1,34 +1,40 @@
...
-    case counter, gauge, histogram, summary, info
+    /// See `PromCounter`
+    case counter
+    /// See `PromGauge`
+    case gauge
+    /// See `PromHistogram`
+    case histogram
+    /// See `PromSummary`
+    case summary
 }
```
ktoso: 👍

Sources/Prometheus/MetricTypes/Metric.swift
```diff
@@ -1,34 +1,40 @@
...
 
 /// Metric protocol
 ///
 /// See https://prometheus.io/docs/concepts/metric_types/
 public protocol Metric {
```
ktoso: I think this one should follow the Prom prefixing I think, so `PromMetric`

Sources/Prometheus/MetricTypes/Metric.swift
```diff
@@ -1,34 +1,40 @@
 /// Different types of metrics supported by SwiftPrometheus
 public enum MetricType: String {
```
ktoso: Should also follow the Prom prefixing, so `PromMetricType`

Sources/PrometheusMetrics/PrometheusMetrics.swift
```diff
@@ -0,0 +1,307 @@
...
+        let createHandler = { (counter: PromCounter) -> CounterHandler in
+            return MetricsCounter(counter: counter, dimensions: dimensions)
+        }
+        if let counter: PromCounter<Int64, DimensionLabels> = self.getMetricInstance(with: label, andType: .counter) {
+            return createHandler(counter)
+        }
+        return createHandler(self.createCounter(forType: Int64.self, named: label, withLabelType: DimensionLabels.self))
+    }
+    
+    /// Makes a recorder
```
ktoso: no need for the comment, super type has a good comment

Sources/PrometheusMetrics/PrometheusMetrics.swift
```diff
@@ -0,0 +1,307 @@
...
+            self.removeMetric(handler.histogram)
+        }
+    }
+    
+    public func destroyTimer(_ handler: TimerHandler) {
+        guard let handler = handler as? MetricsSummary else { return }
+        self.removeMetric(handler.summary)
+    }
+    
+    /// Makes a counter
```
ktoso: no need for the comment, super type has a good comment

Sources/Prometheus/MetricTypes/Counter.swift
```diff
@@ -1,9 +1,11 @@
+import NIOConcurrencyHelpers
+
 /// Prometheus Counter metric
 ///
 /// See https://prometheus.io/docs/concepts/metric_types/#counter
```
ktoso: Is `- See: https://prometheus.io/docs/concepts/metric_types/#counter` better perhaps?
If yes, perhaps apply on all other See links?

Sources/Prometheus/Prometheus.swift
```diff
@@ -45,13 +64,17 @@ public class PrometheusClient {
...
+        withLabelType labelType: U.Type) -> PromCounter<T, U>
     {
-        let counter = Counter<T, U>(name, helpText, initialValue, self)
-        prometheusQueue.async(flags: .barrier) {
+        return self.lock.withLock {
+            if let type = metricTypeMap[name] {
+                precondition(type == .counter, "Label \(name) was associated with \(type) before. Can not be used for a counter now.")
+            }
+            let counter = PromCounter<T, U>(name, helpText, initialValue, self)
+            self.metricTypeMap[name] = .counter
```
ktoso: this is good, and note that even though we want to remove the `_type` on the metric type, we do keep it here and the map shall remain as well -- this is great as it powers the protecting of reusing labels 👍

Package.swift
```diff
@@ -9,15 +9,22 @@ let package = Package(
             name: "SwiftPrometheus",
             targets: ["Prometheus"]),
     ],
+    dependencies: [
+        .package(url: "https://github.com/apple/swift-metrics.git", from: "1.0.0"),
+        .package(url: "https://github.com/apple/swift-nio.git", from: "1.0.0"),
```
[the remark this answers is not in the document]
weissi: Yes, we'd definitely need this to be NIO 2 so it's compatible with the other bits of the SSWG ecosystem. But the migration should really be rather straightforward.

Sources/Prometheus/MetricTypes/Counter.swift
```diff
@@ -35,14 +40,15 @@ public class Counter<NumType: Numeric, Labels: MetricLabels>: Metric, Prometheus
...
+        self.lock = Lock()
     }
     
     /// Gets the metric string for this counter
     ///
     /// - Returns:
     ///     Newline seperated Prometheus formatted metric string
-    public func getMetric(_ done: @escaping (String) -> Void) {
-        prometheusQueue.async(flags: .barrier) {
+    public func getMetric() -> String {
```
ktoso: I'm somewhat worried about this being `getMetric`, can we call it some `renderMetric` perhaps?
Getting does not imply that we'll start marshalling everything into the "wire format" which happens to be the strings...

It gets esp. worrying once we hit APIs like these which create multi-line strings. Would the `render` prefix be okey here you think @MrLotU ? 

Maybe @weissi has another naming idea.

Sources/Prometheus/MetricTypes/Histogram.swift
```diff
@@ -1,8 +1,11 @@
+import NIOConcurrencyHelpers
+
 /// Default buckets used by Histograms
 public var defaultBuckets = [0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1.0, 2.5, 5.0, 7.5, 10.0, Double.greatestFiniteMagnitude]
```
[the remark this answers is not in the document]
weissi: Agree with @ktoso . Also we shouldn't have public global variables outside of any scope. You can nest them into something. If you don't have anything to nest them into, just make a new thing like so:

```swift
enum Prometheus {
    public static let defaultBuckets = [0.005, ...]
}
```

That enum doesn't have any cases so it's purely a namespace. And yes, deffo needs to be a `let`

Sources/PrometheusMetrics/Exports.swift
weissi: Yes, please don't do any `@_exported import`s from stuff that's not in your package. Even inside of your package only do that if you really need it.

weissi: @MrLotU why do you even want to `Prometheus`? Isn't that the module that contains the private implementation? And `CoreMetrics` should definitely not be exported by your lib.

Sources/PrometheusMetrics/PrometheusMetrics.swift
weissi: I think it's for Jazzy so it doesn't show something low like "20% documented". But I haven't seen a comment that explains the conformance, so I'd replace that comment too.

If you want to link to the description of some other thing I think you can use `@see OtherThing.stringValue`

MrLotU:
Thanks a lot @ktoso! I’ll take a look ASAP and address your comments!

circle.yml
```diff
@@ -0,0 +1,43 @@
+version: 2
+
+jobs:
+  swift4:
+    docker:
+      - image: swift:4.2
```
weissi: make that `swift:5.0`, NIO 1 just fine with Swift 5. But it should become NIO2 anyway.

docs/Classes.html
```diff
@@ -21,9 +21,9 @@
     <header class="header">
```
weissi: do we really need `docs/` being checked in on the `master` branch? That'll make all diffs quite huge, right?

Sources/Prometheus/MetricTypes/Gauge.swift
weissi: +1

Sources/Prometheus/MetricTypes/Histogram.swift
```diff
@@ -123,32 +131,28 @@ public class Histogram<NumType: DoubleRepresentable, Labels: HistogramLabels>: M
...
-            return completion()
         }
     }
+
 }
 
 extension PrometheusClient {
-    fileprivate func getOrCreateHistogram<T: Numeric, U: HistogramLabels>(with labels: U, for his: Histogram<T, U>) -> Histogram<T, U> {
+    /// Helper for histograms & labels
+    fileprivate func getOrCreateHistogram<T: Numeric, U: HistogramLabels>(with labels: U, for his: PromHistogram<T, U>) -> PromHistogram<T, U> {
```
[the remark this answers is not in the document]
weissi: I think we should stay clear of almost all abbreviations really, so `histogram` should right :)

Sources/Prometheus/MetricTypes/Metric.swift
weissi: why don't we do `PrometheusMetricType`?

Sources/Prometheus/MetricTypes/Metric.swift
```diff
@@ -1,34 +1,40 @@
...
 /// metrics
 internal protocol PrometheusHandled {
-    var prometheus: PrometheusClient { get }
+    /// Promtheus client handeling this metric
+    var prometheus: PrometheusClient? { get }
 }
 
 /// Base MetricLabels protocol
-public protocol MetricLabels: Codable, Hashable {
+public protocol MetricLabels: Encodable, Hashable {
```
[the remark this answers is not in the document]
weissi: +1, if at possible, I would try to not add any `public protocol`s.

Sources/Prometheus/MetricTypes/Summary.swift
weissi: +1

Sources/Prometheus/Prometheus.swift
weissi: you're right (since Swift 3 or so they're lower-case)

Sources/PrometheusMetrics/Exports.swift
weissi: definitely avoid `@_exported`. If you `@_exported import` for a another package, it breaks SemVer. And even for your own modules it's not good because it pollutes the global namespace.

Package.swift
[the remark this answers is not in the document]
MrLotU: Yeah, I just use the locks from concurrency helpers, so that's all good. Will do

README.md
```diff
@@ -115,6 +126,6 @@ Here, I used [Vapor](https://github.com/vapor/vapor) syntax, but this will work
 
 All contributions are most welcome!
 
-If you think of some cool new feature that should be included, please [create an issue](https://github.com/MrLotU/SwiftPrometheus/issues/new/choose). Or, if you want to implement it yourself, [fork this repo](https://github.com/MrLotU/SwiftPrometheus/fork) and submit a PR!
+-If you think of some cool new feature that should be included, please [create an issue](https://github.com/MrLotU/SwiftPrometheus/issues/new/choose). Or, if you want to implement it yourself, [fork this repo](https://github.com/MrLotU/SwiftPrometheus/fork) and submit a PR!
```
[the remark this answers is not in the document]
MrLotU: Euhm, I think I copied it from a git diff 😅 (no). I'll remove it

Sources/Prometheus/MetricTypes/Histogram.swift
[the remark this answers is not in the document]
MrLotU: I didn't really worry about the wording in this function since it's for internal use, but I'll update it :)

Sources/Prometheus/MetricTypes/Gauge.swift
MrLotU: I might be off on my cycles here, but it's not about not being able to drop the PrometheusClient, but about not being able to drop the Gauge if it holds on to it's client, and the client holds on to the Gauge (it's in it's `metrics` array)

Sources/Prometheus/Utils.swift
MrLotU: Same, hahahaha

Sources/PrometheusMetrics/Exports.swift
MrLotU: 👍 I'll remove them

MrLotU: Also, `Prometheus` is not a private implementation. it's the standalone implementation. The `PrometheusMetrics` implementation is the wrapper based on the `swift-metrics` API

Sources/PrometheusMetrics/PrometheusMetrics.swift
MrLotU: It was indeed for Jazzy. But I can take another look at all comments and give them a bit more meaningful comments 😄

Sources/Prometheus/MetricTypes/Histogram.swift
MrLotU: Yup, requirement set forth by prometheus. The value in the label HAS to be called `le` in Histograms

Sources/Prometheus/MetricTypes/Metric.swift
MrLotU: Sounds good! 👍

Sources/Prometheus/MetricTypes/Counter.swift
MrLotU: Don't think that does anything?

circle.yml
MrLotU: I already have Swift 5 in there too, this was more for backwards compatability to 4.2. But if this is NIO 2, that will not be there anyway, right?

docs/Classes.html
MrLotU: Euhm, don't think so. I'll look into maybe having the docs on a separate branch or somewhere else entirely.

Sources/Prometheus/MetricTypes/Metric.swift
```diff
@@ -1,34 +1,40 @@
...
 /// Metric protocol
 ///
 /// See https://prometheus.io/docs/concepts/metric_types/
 public protocol Metric {
+    /// Name of the metric
     var name: String { get }
+    /// Optional help of the metric
     var help: String? { get }
+    /// Type of the metric
     var _type: MetricType { get }
```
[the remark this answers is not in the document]
MrLotU: It's also used in a few other places and was not specifically introduced for this, but we can have a discussion on this :)

Package.swift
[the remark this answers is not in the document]
weissi: @MrLotU okay, that's one of the rare cases where you could even do 

```
.package(url: "https://github.com/apple/swift-nio.git", Version("1.0.0") ..< Version("3.0.0")),
```

which will give you any NIO 1.x or 2.x, so you'd even be compatible with say Vapor 3 & 4.

circle.yml
ktoso: Correct, NIO2 is only 5+

docs/Classes.html
ktoso: docs branch is usually good, and you can self host the docs pages on gh pages then :) let us know if you need hints about is

Sources/Prometheus/MetricTypes/Metric.swift
ktoso: `Prom` is kind of "well known short name" for prometheus weirdly enough but yeah we could do full names everywhere if we'd want to hm... consistency is king tho :)

Sources/Prometheus/MetricTypes/Histogram.swift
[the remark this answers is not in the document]
ktoso: hah yeah sorry, I reviewed it "all" :)

Sources/Prometheus/MetricTypes/Metric.swift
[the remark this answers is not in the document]
ktoso: Ok, lets have a look at it, what's the challenges you think?

Package.swift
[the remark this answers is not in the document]
ktoso: Oh nice 👍

Sources/Prometheus/Prometheus.swift
MrLotU: Updated 👌

Sources/Prometheus/MetricTypes/Metric.swift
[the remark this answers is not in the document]
MrLotU: Not sure if there are challenges. Just need to think about implementation details. I'll get back on this one 😄

docs/Classes.html
MrLotU: I'll look into updating this 👍

Sources/Prometheus/MetricTypes/Counter.swift
ktoso: Based on https://forums.swift.org/t/discussion-swift-prometheus-implementation/25544 seems like `collect()` would be a good name for those :) Would be good to change before merging I think, the "get" wording is quite weird and it is public API

README.md
```diff
@@ -101,12 +116,8 @@ To keep SwiftPrometheus as clean and lightweight as possible, there is no way of
...
 This could look something like this:
 ```swift
-router.get("/metrics") { request -> Future<String> in
-    let promise = req.eventLoop.newPromise(String.self)
-    prom.getMetrics {
-        promise.succeed(result: $0)
-    }
-    return promise.futureResult
+router.get("/metrics") { request -> String in
+    return myProm.getMetrics()
```
ktoso: don't miss to change the name here when renaming

docs/Classes.html
MrLotU: I removed the docs for now. I'll add them back in on the /docs branch :)

circle.yml
MrLotU: I went with the suggestion from @weissi to use `1.0 ..< 3.0` for NIO :)

Sources/Prometheus/MetricTypes/Counter.swift
MrLotU: Updated it to `.collect()` :)

README.md
```diff
@@ -117,7 +117,7 @@ To keep SwiftPrometheus as clean and lightweight as possible, there is no way of
 This could look something like this:
 ```swift
 router.get("/metrics") { request -> String in
-    return myProm.getMetrics()
+    return myProm.collect()
```
ktoso: excellent!

Yasumoto:
Also very belated 👍 , thank you @MrLotU !! (And thanks @ktoso + @weissi for the reviews!)

Issue #5, opened 2019-07-05
Packaging discussion


ktoso:
So with the `@_exports` gone, the following is not true anymore right?

```
import PrometheusMetrics // Auto imports Prometheus too, but adds the swift-metrics compatibility
```

?

ktoso:
I'm wondering if there is a reason to have the two in separate modules really btw...
Does not really cost much to have just `Prometheus` and it also implement the `Metrics` it is "shared currency" API after all, wdyt @weissi

// Finding those as I'm trying to use it today btw :-)

MrLotU:
Fixed in https://github.com/MrLotU/SwiftPrometheus/commit/31d6964f7240125b07235a1f75e2601dbea74680. Though I'm still not sure how to handle this. I know for a fact that Vapor uses `@_exported` a lot. I also think that in this specific usecase, it doesn't have many of the drawbacks it normally gives you.

The reason I separated the two is because I don't want to pressure people into using the `swift-metrics` APIs becuase in my case, the standalone Prometheus APIs give you more power.

Curious to hear input on this though. CC @weissi

weissi:
Usually, I'd say I agree with @ktoso that one module would probably be better. But the point @MrLotU makes sense (that there might be users who need more power than the 'currency API' gives you).

Regarding the `@_exported import`: What exactly are the things that you would need to have exported? I thought unless you're a special prometheus power-user you'd just live off the API that SwiftMetrics gives you?

So what I was thinking (please correct me if I'm wrong), you'd do in your `main.swift`

```swift
import Metrics
import PrometheusMetrics

PrometheusMetrics.bootstrap(....)
```

but in all other files, you'd just do

```swift
import Metrics

// use the metrics
```

So rephrasing the question: If you are the "normal" case and you want to use SwiftMetrics with Prometheus as a backend, why would you ever want any of the symbols from `Prometheus`?

ktoso:
> Usually, I'd say I agree with @ktoso that one module would probably be better. But the point @MrLotU makes sense (that there might be users who need more power than the 'currency API' gives you).

There seems to be a misunderstanding here I think?

It can be:
- 1 module
- swift-metrics handler implementations
- the PromCounter et al types
- users pick whichever they 

And users should:
- use the powerful apis if they want, sure!
- do bootstrap the  

> If you are the "normal" case and you want to use SwiftMetrics with Prometheus as a backend, why would you ever want any of the symbols from Prometheus?

[...]

PR #6, opened 2019-07-05
 package, also expose PrometheusMetrics

ktoso:
Unless I'm missing something, we should also expose the `PrometheusMetrics` in the product right?

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -7,7 +7,10 @@ let package = Package(
     products: [
         .library(
             name: "SwiftPrometheus",
-            targets: ["Prometheus"]),
+            targets: [
+                "Prometheus",
+                "PrometheusMetrics"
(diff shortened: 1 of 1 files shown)


MrLotU:
I think we do, yes. Though it might make more sense to even split them up into 2 targets. In that case (If I understand how SPM works correctly) you will not have to link & compile swift-metrics if you decide not to use it.

MrLotU:
Much like how NIO has their libraries setup:
https://github.com/apple/swift-nio/blob/master/Package.swift#L94

ktoso:
Relates to https://github.com/MrLotU/SwiftPrometheus/issues/5

ktoso:
Cloding, ongoing discussion seems to point towards other packaging style :)

PR #9, opened 2019-07-10
👯‍♀️ Merge Prometheus and PrometheusMetrics targets into one library

### LINKED ISSUE #5: Packaging discussion

Yasumoto:
This is following discussion in #5.

As a library author and application developer, I'd find this much simpler to use if I'm only working with one library since I expect this functionality to be present.

Very open-minded to improvements or suggested tweaks, however!

### Checklist

### Motivation and Context
We want to provide a simple solution for end-users to take advantage of. `swift-metrics` has broad applicability across the community, and is already included in the dependency graph. If a user does not wish to use `swift-metrics` they don't need to, but this change will make it much clearer how to do so.

This fixes #5.

### Description

I moved `PrometheusMetrics` into the `Prometheus` target, which composes `SwiftPrometheus`.

In addition, I created a `PrometheusExample` executable to make it easier for users to see the output format and follow along with the code (particularly from within Xcode 🔨 ).

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,45 +6,51 @@ A prometheus client for Swift supporting counters, gauges, histograms, summaries
 
 # Usage
 
-For examples, see [main.swift](./Sources/PrometheusExample/main.swift)
+To see a working demo, see [PrometheusExample](./Sources/PrometheusExample/main.swift).
 
 First, we have to create an instance of our `PrometheusClient`:
+
 ```swift
 import Prometheus
 let myProm = PrometheusClient()
 ```
 
-## Usage with Swift-Metrics
-_For more details about swift-metrics, check the GitHub repo [here](https://github.com/apple/swift-metrics)_
+## Usage with SwiftMetrics
+_For more details about swift-metrics, please view [the GitHub repo](https://github.com/apple/swift-metrics)._
 
-To use SwiftPrometheus with swift-metrics, all the setup required is this:
+To use SwiftPrometheus with swift-metrics, you need to configure the backend inside the `MetricsSystem`:
+
 ```swift
+import Metrics
 import Prometheus
-import PrometheusMetrics
 let myProm = PrometheusClient()
 MetricsSystem.bootstrap(myProm)
 ```
 
-To use prometheus specific features in a later stage of your program, or to get your metrics out of the system, there is a convenience method added to `MetricsSystem`:
+To use prometheus-specific features in a later stage of your program, or to get your metrics out of the system, there is a convenience method added to `MetricsSystem`:
+
 ```swift
-// This is the same instance was used in `.bootstrap()` earlier.
+// This returns the same instance passed in to `.bootstrap()` earlier.
 let promInstance = try MetricsSystem.prometheus()
+print(promInstance.collect())
 ```
-You can than use the same APIs that are layed out in the rest of this README
+
+You can then use the same APIs described in the rest of this README.
 
 ## Counter
 
-Counters go up, and reset when the process restarts.
+Counters go up (they can only increase in value), and reset when the process restarts.
 
 ```swift
 let counter = myProm.createCounter(forType: Int.self, named: "my_counter")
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -6,45 +6,51 @@ A prometheus client for Swift supporting counters, gauges, histograms, summaries
...
 
-Counters go up, and reset when the process restarts.
+Counters go up (they can only increase in value), and reset when the process restarts.
 
 ```swift
 let counter = myProm.createCounter(forType: Int.self, named: "my_counter")
 counter.inc() // Increment by 1
-counter.inc(12) // Increment by given value 
+counter.inc(12) // Increment by given value
+counter.reset() // Set the value back to 0
```
MrLotU: We don't have a reset on counters :)

README.md
```diff
@@ -73,16 +79,16 @@ summary.observe(4.7) // Observe the given value
 
 ## Info
 
-Info tracks key-value information, usually about a whole target.
+Info tracks key-value information, usually about a whole target. These are typically helpful details like git sha or other metadata.
```
MrLotU: This entire section should go, it's not in the source anymore (that's my bad)

MrLotU: If you don't want to include it here, I'll create a separate commit/PR to remove this.

README.md
```diff
@@ -113,15 +120,15 @@ counter.inc(12, .init(route: "/"))
 
 # Exporting
 
-To keep SwiftPrometheus as clean and lightweight as possible, there is no way of exporting metrics directly to Prometheus. Instead, retrieve a formatted string that Prometheus can use, so you can integrate it in your own Serverside Swift application
+Prometheus itself is designed to "pull" metrics from a destination. Following this pattern, SwiftPrometheus is designed to expose metrics, as opposed to submitted/exporting them directly to Prometheus itself. Instead, SwiftPrometheus produces a formatted string that Prometheus can parse, which can be integrated into your own application.
```
MrLotU: I think you can remove the word `Instead,` here.
```suggestion
Prometheus itself is designed to "pull" metrics from a destination. Following this pattern, SwiftPrometheus is designed to expose metrics, as opposed to submitted/exporting them directly to Prometheus itself. SwiftPrometheus produces a formatted string that Prometheus can parse, which can be integrated into your own application.
```

README.md
Yasumoto: You're totally right! From the [prom doc about Counters](https://prometheus.io/docs/concepts/metric_types/#counter):

> or be reset to zero on restart.

And [from Brian](https://www.robustperception.io/how-does-a-prometheus-counter-work):

> Any time a counter appears to decrease it'll be treated as though there was a reset to 0 right after the first data point. This makes it important that it not be possible for Counters to be decremented, a Counter that has the potential to be decremented is in reality a Gauge.

So given that [this is part of the `Counter` class](https://github.com/apple/swift-metrics/blob/master/Sources/CoreMetrics/Metrics.swift#L80) (which is what threw me off initially 😂 ) maybe we should do something different [in the code](https://github.com/MrLotU/SwiftPrometheus/blob/master/Sources/PrometheusMetrics/PrometheusMetrics.swift#L21) like throw an error (unlikely) or log a warning?

README.md
Yasumoto: Great eyes, thanks!

README.md
MrLotU: > maybe we should do something different in the code like throw an error (unlikely) or log a warning?

Not quite sure what to do here. I think throwing an error would be best, but the API does not allow that. Logging is not ideal either IMO since that means we'd have to take `swift-log` as a dependency. I think I'll open an issue about it on the https://github.com/apple/swift-metrics repo 😄 For now. This looks great!

MrLotU:
@Yasumoto Thanks for this PR!

PR #12, opened 2019-07-23
Get metric promises

MrLotU:
CC @Lukasa @weissi @ktoso

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -14,7 +14,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-metrics.git", from: "1.0.0"),
-        .package(url: "https://github.com/apple/swift-nio.git", Version("1.0.0") ..< Version("3.0.0")),
+        .package(url: "https://github.com/apple/swift-nio.git", from: "2.0.0"),
     ],
     targets: [
         .target(
--- a/Package.swift
+++ b/Package.swift
@@ -19,7 +19,7 @@ let package = Package(
     targets: [
         .target(
             name: "Prometheus",
-            dependencies: ["CoreMetrics", "NIOConcurrencyHelpers"]),
+            dependencies: ["CoreMetrics", "NIOConcurrencyHelpers", "NIO"]),
         .target(
             name: "PrometheusExample",
             dependencies: ["Prometheus", "Metrics"]),
--- a/Sources/Prometheus/MetricTypes/PromMetric.swift
+++ b/Sources/Prometheus/MetricTypes/PromMetric.swift
@@ -34,8 +34,12 @@ public protocol PromMetric {
     
     /// Retrieves the Prometheus-formatted metric data
     func collect() -> String
-    
-    func collect(into buffer: inout ByteBuffer)
+}
+
+extension PromMetric {
+    func collect(into buffer: inout ByteBuffer) {
+        buffer.writeString(collect())
+    }
 }
 
 /// Adding a prometheus instance to all metrics
--- a/Sources/Prometheus/MetricTypes/PromMetric.swift
+++ b/Sources/Prometheus/MetricTypes/PromMetric.swift
@@ -1,3 +1,5 @@
+import NIO
+
 /// Different types of metrics supported by SwiftPrometheus
 public enum PromMetricType: String {
     /// See `PromCounter`
@@ -32,6 +34,8 @@ public protocol PromMetric {
     
     /// Retrieves the Prometheus-formatted metric data
     func collect() -> String
+    
+    func collect(into buffer: inout ByteBuffer)
 }
 
 /// Adding a prometheus instance to all metrics
--- a/Sources/Prometheus/MetricTypes/PromMetric.swift
+++ b/Sources/Prometheus/MetricTypes/PromMetric.swift
@@ -37,6 +37,10 @@ public protocol PromMetric {
 }
 
 extension PromMetric {
+    /// Helper method to record metrics into a `ByteBuffer` directly
+    ///
+    /// - Parameters:
+    ///     - buffer: `ByteBuffer` to collect into
     func collect(into buffer: inout ByteBuffer) {
         buffer.writeString(collect())
     }
--- a/Sources/Prometheus/Prometheus.swift
+++ b/Sources/Prometheus/Prometheus.swift
@@ -27,9 +27,9 @@ public class PrometheusClient {
     ///
     /// - Parameters:
     ///     - p: Promise that will succeed with a newline separated string with metrics for all Metrics this PrometheusClient handles
-    public func collect(_ p: EventLoopPromise<String>) {
+    public func collect(_ succeed: (String) -> ()) {
         self.lock.withLock {
-            p.succeed(self.metrics.map { $0.collect() }.joined(separator: "\n"))
+            succeed(self.metrics.map { $0.collect() }.joined(separator: "\n"))
         }
     }
     
@@ -37,12 +37,15 @@ public class PrometheusClient {
     ///
     /// - Parameters:
     ///     - p: Promise that will succeed with a newline separated string with metrics for all Metrics this PrometheusClient handles
-//    public func collect(_ p: EventLoopPromise<ByteBuffer>) {
-//        self.lock.withLock {
-//            let buffer = ByteBuffer()
-////            p.succeed(result: self.metrics.map { $0.collect() }.joined(separator: "\n"))
-//        }
-//    }
+    public func collect(_ succeed: (ByteBuffer) -> ()) {
+        self.lock.withLock {
+            var buffer = ByteBufferAllocator().buffer(capacity: 0)
+            self.metrics.forEach {
+                $0.collect(into: &buffer)
+            }
+            succeed(buffer)
+        }
+    }
     
     // MARK: - Metric Access
     
--- a/Sources/Prometheus/Prometheus.swift
+++ b/Sources/Prometheus/Prometheus.swift
@@ -23,10 +23,12 @@ public class PrometheusClient {
         self.lock = Lock()
     }
     
+    // MARK: - Collection
+    
     /// Creates prometheus formatted metrics
     ///
     /// - Parameters:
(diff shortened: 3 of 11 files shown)


Package.swift
```diff
@@ -19,7 +19,7 @@ let package = Package(
     targets: [
```
weissi: unfortunately, you'll now need to change the NIO dependency to `from: "2.0.0"` as it won't work with NIO 1

Sources/Prometheus/Prometheus.swift
```diff
@@ -24,10 +25,11 @@ public class PrometheusClient {
...
-    /// - Returns: Newline separated string with metrics for all Metric Trackers of this Prometheus instance
-    public func collect() -> String {
-        return self.lock.withLock {
-            return self.metrics.map { $0.collect() }.joined(separator: "\n")
+    /// - Parameters:
+    ///     - p: Promise that will succeed with a newline separated string with metrics for all Metrics this PrometheusClient handles
+    public func collect(_ p: EventLoopPromise<String>) {
+        self.lock.withLock {
+            p.succeed(result: self.metrics.map { $0.collect() }.joined(separator: "\n"))
         }
```
ktoso: I would agree with @Lukasa that "both are useful" from https://forums.swift.org/t/feedback-swift-prometheus-implementation/26928

>  Naturally having access to a String version is convenient and I don't think we should throw it away, but having both is likely helpful as well.

As I've personally "abused" the synchronous one a little bit to get "dump me all the things, I don't have an event loop and I don't care" in small toy apps. Could we keep both APIs and add a node that the blocking one should be avoided?

Maybe a note like `/// This call is BLOCKING and not intended for production use, use `collect(promise:)` instead?

WDTY @Lukasa?

ktoso: OR are we afraid that people'd abuse that call? I mean, they could always to the promise and block on it... mixed feelings if we should offer the `-> String` both answers are a bit meh :-\ Would be a no brainer if there was a "GlobalExecutionContext" but we don't have one in Swift...

Package.swift
MrLotU: :(

Sources/Prometheus/Prometheus.swift
MrLotU: I do indeed think this will be abused. Even though it is nice for testing apps. But than still, setting up an EventLoop and creating a promise can be done in 1 line of code, so I don't think it's all that bad :)

Package.swift
MrLotU: Is that the case though? Because does NIO 2 and NIO 3 not both use the same methods on `EventLoopPromise` to succeed it? I think just the promise creation changed, no?

MrLotU: https://circleci.com/gh/MrLotU/SwiftPrometheus/86?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link ran & succeeded with NIO 1.14.1

MrLotU: Nevermind, not true. It only works on NIO 1.x now. I'll push an update :)

Sources/Prometheus/Prometheus.swift
tomerd: another option is to use a offer a callback based signature eg `collect(_ callback: (String) -> Void)`

Lukasa: I think a callback based approach would be good too.

Lukasa: Maybe let's do the callback instead of the promise? Then folks can wrap this to do whatever they want.

weissi: but the callback should yield an `Result<String, Error>` or `Result<ByteBuffer, Error>` IMHO, just like the synchronous version (if we decide to have one) should be `throws -> String` / `throws -> ByteBuffer`.

weissi: or can it truly never fail?

MrLotU: Right now it can not fail. It's not throwing ATM either.

MrLotU: Converting it to a callback as we speak btw :)

weissi: where does the callback get called on actually? Should it take a `queue: DispatchQueue` or what's it executing on?

Lukasa: @weissi The simplest version is the "don't ask any questions, you make sure you go where you need to". That's maximally compatible with non-dispatch-based systems, but the most annoying for Dispatch-based ones. We could pass a dispatch queue if we want.

Package.swift
```diff
@@ -14,12 +14,12 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-metrics.git", from: "1.0.0"),
-        .package(url: "https://github.com/apple/swift-nio.git", Version("1.0.0") ..< Version("3.0.0")),
+        .package(url: "https://github.com/apple/swift-nio.git", from: "2.0.0"),
```
Yasumoto: Can we keep this from `1.0.0` to still work with `Vapor 3` 💧 until 4 is released?

Sources/Prometheus/Prometheus.swift
Yasumoto: So much of the ecosystem is using NIO-based promises that it's nice to maintain that pattern. However, using the stdlib's `Result` and a callback would be nice.

I'm not a fan of imposing limits only through comments (regarding `/// This call is BLOCKING and not intended for production use,`) since it's almost too easy to miss that and think you have the "easy" option available.

Package.swift
Yasumoto: Can we keep this from 1.0.0 to still work with Vapor 3 💧 until 4 is released?

Maybe we can/should add compiler directives so we could work with both?

weissi: @Yasumoto / @MrLotU Couldn't you tag version `0.0.0` for NIO1 support and then do `1.0.0` for NIO2?

MrLotU: If there is no way to differentiate in code between NIO 1 and 2, than yes, I think that's the way to go.

Sources/Prometheus/Prometheus.swift
MrLotU: I'm OK with either. It's a callback right now, though using the `Result` has no real added value to me since it can never fail right now. If that ever changes it would be a SemVer Major anyways (since it'll become throwing) so we can do that later.

MrLotU:
> IMHO the protocol should only have one way (the most versatile one) to do that in the protocol. The convenience functionality can be done using extensions.

Good point. I think I'll add the ByteBuffer one in an extension 😄

Sources/Prometheus/Prometheus.swift
ktoso: That a lot of ecosystem depends on NIO Future/Promise is true, but it's not a happy place IMHO, we'll have to break all these once actual futures land or if we generalise them or whichever happens first. 

I'd be okey with NIO futures/promise since that's status quo and easy to integrate with. I do think it's a bit sad that we have to pull in NIO only for the promises but yeah, no real better option out there nowadays.

The callback is nice actually since we don't leak the NIO types then, I would like that, but I'm also okey with the `ELPromise`.

MrLotU:
> I was proposing that only the high level API should be sync. However, I'd forgotten that these were public. Do they need to be? Is there utility in collecting only one metric?

@Lukasa Not sure if there is much utility. I use them myself for testing, but not sure what people would like to use them for. In any case, collecting 1 metric should be pretty quick.

Sources/Prometheus/Prometheus.swift
MrLotU: @weissi If we ditch the promise, would that mean we can go back to using either NIO 1 or NIO 2 again? Or are there also differences in `ByteBuffer`

weissi: `ByteBuffer`'s API differ quite a lot between 1 and 2.

MrLotU: Well, there goes that hope, haha. 2 branches it is than :) Still doesn't help decide upon either Promise or normal closure though 🙁

ktoso:
So we're converging on not offering the synchronous one at all right? That's fine I think...
Like @MrLotU says, my case where I used the synchronous one is also a toy example, and we don't really need (or should) optimize for toy examples but getting the real systems right -> Promise or callback.

Package.swift
Yasumoto: Great save, let's go with a `0.0.0` release train for NIO1.

Yasumoto:
I'd prefer a `Promise` since it's consistent with other tools/libraries (yes, at the cost of tying closely with NIO) but I'm open-minded to the final solution 👍

Tests/SwiftPrometheusTests/PrometheusMetricsTests.swift
```diff
@@ -1,27 +1,37 @@
 import XCTest
+import NIO
 @testable import Prometheus
 @testable import CoreMetrics
 
 final class PrometheusMetricsTests: XCTestCase {
     
     var prom: PrometheusClient!
+    var group: EventLoopGroup!
+    var elg: EventLoop {
```
Yasumoto:
```suggestion
    var eventLoop: EventLoop {
```

Since it's not an `EventLoopGroup`

MrLotU:
I've added methods to `PrometheusClient` with both `String`s and `ByteBuffers` both in callback & promise form. I figured it'd give users the most freedom to choose which to use.

Would love if you could give this another quick look :)

MrLotU:
Once this is merged, I'll also create a separate branch for NIO 1 support.

Sources/Prometheus/MetricTypes/PromMetric.swift
```diff
@@ -1,3 +1,5 @@
+import NIO
```
ktoso: side topic question: don't all sources have to start with a copyright?

MrLotU: Euhm, I honestly don't know. I don't think they HAVE to, but I'll double check.

ktoso:
AFAICS this is good enough so far... We can take it further when and if the need arises 👍

PR #13, opened 2019-08-28
Do not recreate a new counter if we already have one.

Yasumoto:
### Checklist

### Motivation and Context
We want to make sure we're _reusing_ metrics, not losing ones we already create.

### Description

I discovered when running a service that I'm getting multiple invocations of the same metric appearing when I `collect()`:

```
# TYPE invocation_register counter
invocation_register 0
invocation_register{tool="simple-curl"} 1
# TYPE invocation_register counter
invocation_register 0
invocation_register{tool="simple-curl"} 1
# TYPE invocation_register counter
invocation_register 0
invocation_register{tool="simple-curl"} 1
```

I'm calling [`createCounter`](https://github.com/Yasumoto/emergency-stop/pull/3/files#diff-e579ad4a0b1db41bb8b074fbd630b234R67) on each request based [on the good advice I got here](https://github.com/apple/swift-metrics/issues/38#issuecomment-511760602) from @ktoso.

If this looks roughly right to y'all, I can update the other types to follow this pattern, though definitely open to feedback 🎉

### DIFF
--- a/Sources/Prometheus/MetricTypes/Summary.swift
+++ b/Sources/Prometheus/MetricTypes/Summary.swift
@@ -14,7 +14,7 @@ extension SummaryLabels {
     }
 }
 
-/// Prometheus Counter metric
+/// Prometheus Summary metric
 ///
 /// See https://prometheus.io/docs/concepts/metric_types/#summary
 public class PromSummary<NumType: DoubleRepresentable, Labels: SummaryLabels>: PromMetric, PrometheusHandled {
--- a/Sources/Prometheus/Prometheus.swift
+++ b/Sources/Prometheus/Prometheus.swift
@@ -5,7 +5,7 @@ import NIO
 ///
 /// See https://prometheus.io/docs/introduction/overview/
 public class PrometheusClient {
-    
+
     /// Metrics tracked by this Prometheus instance
     private var metrics: [PromMetric]
     
@@ -109,7 +109,6 @@ public class PrometheusClient {
             if let type = metricTypeMap[name] {
                 precondition(type == .counter, "Label \(name) was associated with \(type) before. Can not be used for a counter now.")
             }
-
             let counter = PromCounter<T, U>(name, helpText, initialValue, self)
             self.metricTypeMap[name] = .counter
             self.metrics.append(counter)
@@ -154,6 +153,10 @@ public class PrometheusClient {
         initialValue: T = 0,
         withLabelType labelType: U.Type) -> PromGauge<T, U>
     {
+        if let gauge: PromGauge<T, U> = getMetricInstance(with: name, andType: .gauge) {
+            return gauge
+        }
+
         return self.lock.withLock {
             if let type = metricTypeMap[name] {
                 precondition(type == .gauge, "Label \(name) was associated with \(type) before. Can not be used for a gauge now.")
@@ -202,6 +205,10 @@ public class PrometheusClient {
         buckets: [Double] = Prometheus.defaultBuckets,
         labels: U.Type) -> PromHistogram<T, U>
     {
+        if let histogram: PromHistogram<T, U> = getMetricInstance(with: name, andType: .histogram) {
+            return histogram
+        }
+
         return self.lock.withLock {
             if let type = metricTypeMap[name] {
                 precondition(type == .histogram, "Label \(name) was associated with \(type) before. Can not be used for a histogram now.")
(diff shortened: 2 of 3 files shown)


MrLotU:
This is by design. It's different between `swift-metrics` and this library, but the idea from the `SwiftPrometheus` standpoint is that you create a counter once, than increment that during your application lifecycle, passing in different labels as required.

Might have to reconsider this though, seeing how `swift-metrics` takes a different approach.

CC @tomerd @ktoso

ktoso:
To clarify: Why are you not using the `SwiftMetrics` API to call through to SwiftPrometheus in your app? It would do the right thing, as it's implemented as:

```
    public func makeCounter(label: String, dimensions: [(String, String)]) -> CounterHandler {
        let createHandler = { (counter: PromCounter) -> CounterHandler in
            return MetricsCounterHandler(counter: counter, dimensions: dimensions)
        }
        // vvvvvvv the "right" thing / "semantics you expected"
        if let counter: PromCounter<Int64, DimensionLabels> = self.getMetricInstance(with: label, andType: .counter) {
            return createHandler(counter)
        }
        // ^^^^^^^ 

        return createHandler(self.createCounter(forType: Int64.self, named: label, withLabelType: DimensionLabels.self))
    }
```

Why not use the metrics API as a fascade there as it was intended? 

This library IS an implementation of the metrics SPI, and thus this works:

```
let c1 = Counter(label: "one")
c1.increment()
let c2 = Counter(label: "one")
c2.increment()
```

why not use it, rather than bind yourself into the prometheus API directly?

----

> Do we want to fully panic & bail out of execution?

No, I don't think so. One cannot compare Go's panic to fatal erroring in Swift. Panics are recoverable, Swift faults are not.

Throwing an error sounds good if the same metrics is attempted to be created.
Though I'm actually surprised a bit -- I thought the Go impls semantics was only to bail out of the registered type did not match, not a "re-creation of correct type" 🤔 

> Should we create some kind of "auto" registration functions that do the panic, while the main ones just throw an error which pulls out the existing metric?

By "auto" you mean "return me the same instance if it existed already"? That's what the SwiftMetrics API would do; What am I missing?

// nitpick, let's not use the word panic, mixes up things a bit IMHO... Swift does not do panics, Go does, and they're recoverable.

Yasumoto:
I think there are valid usecases in an environment where Prometheus is such a standard that using this package directly is fine and has some benefits. For instance, using `createCounter ` means you can specify help text for the metric that'll show up in Prometheus. Totally agreed that in general even in that environment using `SwiftMetrics` will make an eventual migration much easier.

That said, the goal for this issue is to track down the expected behavior when using `SwiftPrometheus` itself. I like the `SwiftMetrics implementation, but I can see why in the golang implementation they wanted to explicitly let folks know in case they're clobbering a metric with another library.

Any opposition to throwing an error that also includes a reference to the existing metric? That way callers can surface the collision just in case, but have access to use it if that's acceptable.

--- 

> No, I don't think so. One cannot compare Go's panic to fatal erroring in Swift. Panics are recoverable, Swift faults are not.

Yikes, you're totally right. I definitely assumed `panic` meant "game over, execution will stop", but definitely not the case. From [the go blog](https://blog.golang.org/defer-panic-and-recover):

> *Recover* is a built-in function that regains control of a panicking goroutine.


> I thought the Go impls semantics was only to bail out of the registered type did not match, not a "re-creation of correct type" 🤔

Me too, I was surprised a bit as well.

> By "auto" you mean "return me the same instance if it existed already"? That's what the SwiftMetrics API would do; What am I missing?

`SwiftMetrics` is a-ok 👌 , I'd like `SwiftPrometheus` to do something similar with its behavior when used directly (see above section for my current preference).

Yasumoto:
I played around with this a bit, and it "feels" better to follow the `SwiftMetrics` style and return the metric if it already exists. Open to other preferences though!

Yasumoto:
@MrLotU @ktoso friendly nudge 😄

tomerd:
> I played around with this a bit, and it "feels" better to follow the SwiftMetrics style and return the metric if it already exists.

in one of the original iteration of swift-metrics, we did that for "free" as part of the API package (caching basically) but we decided to remove it and leave it to the implementation to decide. imo, the expected behavior when type and other argument align is to return the existing one

Yasumoto:
@tomerd ah, that's good context and totally makes sense. Given the behavior of the `golang` prometheus client library (throw an exception if you try to create a metric that already exists) I see the value in leaving it open.

For now I think keeping this simple is fine, but I'm definitely open to other feedback too.

Yasumoto:
Looks like this needed a rebase off master 👍

Yasumoto:
Thanks, folks! 🎊

Yasumoto:
@MrLotU do you mind adding this to the 1.0 milestone? I would but not sure I have the permissions. 🙇‍♂️

MrLotU:
Done. Good catch

PR #16, opened 2019-09-28
Fix missing newline between metrics when collecting #15

### LINKED ISSUE #15: Weird collect()-ed formatting, breaking collection
I'm seeing errors in in collecting metrics using `prom.collect(String)`

### Steps to reproduce

Can't share exact code but will work on reproducer, so far tried to reproduce in tests here but failed, something seems weird.

I got a counter, with some labels and a gauge; Exactly like in the snippet below.

### Expected behavior

Scraping should not break.

### Actual behavior

```

# TYPE thingy counter
thingy 0
thingy{a="aaa", b="x"} 9
thingy{a="bbb", b="x"} 6# TYPE my_gauge gauge
my_gauge 0.0
```

So the  `thingy{a="bbb", b="x"} 6# TYPE my_gauge gauge` is missing a newline and this breaks scraping:

`level=warn ts=2019-09-27T08:39:48.356Z caller=scrape.go:930 component="scrape manager" scrape_pool=local target=http://127.0.0.1:8888/metrics msg="append failed" err="strconv.ParseFloat: parsing \"6#\": invalid syntax"`

### Environment

* Framework version: master branch right now; https://github.com/MrLotU/SwiftPrometheus/commit/05eed571db0d6a9cca0573e1e1ff9423e49bbaf0

I'm serving this with NIO directly: 

```
        prom.collect { (buf: ByteBuffer) in
            //...
            context.write(self.wrapOutboundOut(.body(.byteBuffer(buf))))
```

So pretty sure I've had no chance to break this somehow manually.

ktoso:
Resolves #15

### Motivation and Context

I've tried out the backend in a real project and noticed that collecting breaks -- a missing \n between metrics. See #15

Newlines must be inserted between metrics otherwise parsing fails.

The other implementation of collect does it right, since it does `succeed(self.metrics.map { $0.collect() }.joined(separator: "\n"))`

The trailing newline is not an issue, so let's go with this rather than reinventing joined manually IMHO

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /*.xcodeproj
 build/
 .swiftpm/
+.idea
--- a/Package.resolved
+++ b/Package.resolved
@@ -6,8 +6,8 @@
         "repositoryURL": "https://github.com/apple/swift-metrics.git",
         "state": {
           "branch": null,
-          "revision": "8e5110dcd6584ff5acfec40e31263a8d07cba783",
-          "version": "1.1.0"
+          "revision": "3fefedaaef285830cc98ae80231140122076a7e0",
+          "version": "1.2.0"
         }
       },
       {
@@ -15,8 +15,8 @@
         "repositoryURL": "https://github.com/apple/swift-nio.git",
         "state": {
           "branch": null,
-          "revision": "6a79d4687ede7c088fe3e31d9b3452d6e6cd91eb",
-          "version": "2.5.1"
+          "revision": "9201908b54578aa33f1d1826a5a680aca8991843",
+          "version": "2.8.0"
         }
       }
     ]
--- a/README.md
+++ b/README.md
@@ -121,4 +121,3 @@ All contributions are most welcome!
 If you think of some cool new feature that should be included, please [create an issue](https://github.com/MrLotU/SwiftPrometheus/issues/new/choose). Or, if you want to implement it yourself, [fork this repo](https://github.com/MrLotU/SwiftPrometheus/fork) and submit a PR!
 
 If you find a bug or have issues, please [create an issue](https://github.com/MrLotU/SwiftPrometheus/issues/new/choose) explaining your problems. Please include as much information as possible, so it's easier for me to reproduce (Framework, OS, Swift version, terminal output, etc.)
-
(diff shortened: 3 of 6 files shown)


ktoso:
We don't have a CI for PR validation? 🤔

Yasumoto:
@ktoso do you have time to add some? Otherwise we should think about what we want to go with, maybe try out `GitHub Actions`?

Yasumoto:
(I should have time in the next few days if no one else has cycles)

ktoso:
@tomerd seems this project has no CI, do we want to include it in the https://ci.swiftserver.group perhaps? It's a pretty small/quick build.

Yasumoto:
Oh, that'd be awesome! [Reverse engineering](https://ci.swiftserver.group/job/async-http-client-swift51-prb/3/console) the setup, looks like it's using the [`docker compose`](https://github.com/swift-server/async-http-client/blob/master/docker/docker-compose.yaml) configs to run the tests with the right setup?

ktoso:
Yeah correct, we use docker to have a well defined and "run the same env" locally "on linux" (on mac on docker) when testing locally as on CI.

@tomerd maintains that CI so ultimately his call, but I think it'd be nice to be able to offer it to projects if they don't have their own setup. OR we set up travis / circle / githubactions or anything here. I'm happy to help on that once we decide that we need.

Yasumoto:
Cool, I'm good with any solution overall, but might be nice to start with an "externally accessible" build/CI system so projects that want to move into the SSWG can start-then-continue with one of travis/circle ci/github. Otherwise there might be some pressure to migrate into this blessed-but-closed CI system.

I might be over thinking it a bit, but enough projects bootstrap via cargo-culting that allowing new projects to copy-paste best practices they can use from existing projects could be nice.

ktoso:
If you want to set up CI in any other way here, please do ;-) just mentioning the option exists. No strong feelings either way

Yasumoto:
(Your reply was very nice! Thank you so much for offering; I’m also noodling on big-picture ideas but like leveraging maintained infra other SSWG projects use in general. Thanks for talking it through with me!)

MrLotU:
I used to have circle set up for this repo, not sure where that went off to. I'll investigate. @ktoso thanks for the PR!

MrLotU:
Found out it was me who had a setting wrong in CircleCI regarding forked PRs. I'll push something to this branch to trigger a build

ktoso:
Cool, we have Circle then, thank you @MrLotU 🙏

Yasumoto:
Sweet! Thanks @MrLotU & @ktoso 🎉

ktoso:
Yay, great to have CI back, thanks @MrLotU

PR #25, opened 2019-11-09
Get Histogram up to Prometheus spec

### LINKED ISSUE #24: Get Histogram up to Prometheus spec
The parts that are missing:

- [x] Custom bucketing functions (`linear` and `exponential`)
- [x] `time()` -> Some way to time code for users in seconds

See: https://prometheus.io/docs/instrumenting/writing_clientlibs/#histogram

MrLotU:
Fixes #24 

### Checklist
- [x] Custom bucketing functions (`linear` and `exponential`)
- [x] `time()` -> Some way to time code for users in seconds

See: https://prometheus.io/docs/instrumenting/writing_clientlibs/#histogram


MrLotU:
Nudge @ktoso. Would you be able to take a look at this? Especially this part: https://github.com/MrLotU/SwiftPrometheus/pull/25/files#diff-9f9ae2482df3cc6983b73ef1bcc0c7c9R14-R27

ktoso:
Thanks for the ping, will do today!

Sources/Prometheus/MetricTypes/Histogram.swift
```diff
@@ -1,4 +1,73 @@
...
+            self.buckets = Buckets.defaultBuckets.buckets
+            return
+        }
+        var r = r
+        if !r.contains(Double.greatestFiniteMagnitude) {
+            r.append(Double.greatestFiniteMagnitude)
+        }
+        // TODO: Decide if we want to fix these "programmer mistakes" or assert them.
+        r.sort(by: <)
+        assert(Array(Set(r)).sorted(by: <) == r, "Buckets contain duplicate values.")
```
ktoso: I see... I'd somewhat expect people putting in custom buckets not so often -- and even if, they'd probably declare a constant somewhere rather than make them up dynamically. In that sense, perhaps it is better to crash hard here, since it's a static configuration most likely.

We should document that and explain what we do.

In other words -- i'd be fine with crashing and perhaps even this could be an  precondition?

Sources/Prometheus/MetricTypes/Histogram.swift
```diff
@@ -144,7 +213,21 @@ public class PromHistogram<NumType: DoubleRepresentable, Labels: HistogramLabels
...
     }
-
+    
+    /// Time the duration of a closure and observe the resulting time in seconds.
+    ///
+    /// - parameters:
+    ///     - labels: Labels to attach to the resulting value.
+    ///     - body: Closure to run & record.
+    @inlinable
+    public func time<T>(_ labels: Labels? = nil, _ body: @escaping () throws -> T) rethrows -> T {
```
ktoso: Not sure we really need this but if you want to include it then ok

Tests/SwiftPrometheusTests/HistogramTests.swift
```diff
@@ -0,0 +1,120 @@
...
+
+        histogram.observe(1)
+        histogram.observe(2)
+        histogramTwo.observe(3)
+        
+        histogram.observe(3, .init(myValue: "labels"))
+
+        XCTAssertEqual(histogram.collect(), "# HELP my_histogram Histogram for testing\n# TYPE my_histogram histogram\nmy_histogram_bucket{myValue=\"*\", le=\"0.5\"} 0.0\nmy_histogram_bucket{myValue=\"*\", le=\"1.0\"} 1.0\nmy_histogram_bucket{myValue=\"*\", le=\"2.0\"} 2.0\nmy_histogram_bucket{myValue=\"*\", le=\"3.0\"} 4.0\nmy_histogram_bucket{myValue=\"*\", le=\"5.0\"} 4.0\nmy_histogram_bucket{myValue=\"*\", le=\"+Inf\"} 4.0\nmy_histogram_count{myValue=\"*\"} 4.0\nmy_histogram_sum{myValue=\"*\"} 9.0\nmy_histogram_bucket{myValue=\"labels\", le=\"0.5\"} 0.0\nmy_histogram_bucket{myValue=\"labels\", le=\"1.0\"} 0.0\nmy_histogram_bucket{myValue=\"labels\", le=\"2.0\"} 0.0\nmy_histogram_bucket{myValue=\"labels\", le=\"3.0\"} 1.0\nmy_histogram_bucket{myValue=\"labels\", le=\"5.0\"} 1.0\nmy_histogram_bucket{myValue=\"labels\", le=\"+Inf\"} 1.0\nmy_histogram_count{myValue=\"labels\"} 1.0\nmy_histogram_sum{myValue=\"labels\"} 3.0")
```
ktoso: perhaps nicer as a 
```
"""
...
...
"""
```

string?

Tests/SwiftPrometheusTests/HistogramTests.swift
```diff
@@ -0,0 +1,120 @@
...
+        my_histogram_sum{myValue="labels"} 3.0
+        """)
+    }
+    
+    func testHistogramTime() {
+        let histogram = prom.createHistogram(forType: Double.self, named: "my_histogram")
+        let delay = 0.05
+        histogram.time {
+            Thread.sleep(forTimeInterval: delay)
+        }
```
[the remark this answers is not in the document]
MrLotU: To avoid weird behaviour, we use a `startsWith` in the assert, so that the exact time does not matter 😄

Sources/Prometheus/MetricTypes/Histogram.swift
MrLotU: Not my choice, requirement set forth by Prometheus 😄

Tests/SwiftPrometheusTests/HistogramTests.swift
MrLotU: Probably a good idea. Yes. Will update

Sources/Prometheus/MetricTypes/Histogram.swift
ktoso: recommended I guess, but yeah now I noticed it's a SHOULD have the methods -- good to add then! Thanks!

Sources/Prometheus/MetricTypes/Histogram.swift
MrLotU: 👍 Time to crash! For now I'll keep these as `asserts`, we can always see in the future if they should become preconditions

PR #28, opened 2019-12-03
Fix vapor example

code28:
(Since I only updated the README, I skip the checks.)

I fixed the example for using SwiftPrometheus in Vapor. Since there is no function `collect()`, which returns a `String`, we need to prepare a promise in order to collect the data.
This is also the way, the `PrometheusExample` prints the metrics.


MrLotU:
Thanks @code28 !

PR #31, opened 2020-03-25
Update SwiftPrometheus to use SwiftMetric 2.* (TimeUnit change to struct)

tdotclare:
Updates SwiftPrometheus to depend on SwiftMetrics 2.0.

### Checklist
- [X] The provided tests still run.
- [X] I've created new tests where needed.

* note - only unit test change is to verify microsecond display works

### Motivation and Context
Change of TimeUnit in SwiftMetrics from enum to struct allows arbitrary addition of time unit scales in future usage.

Removes timeUnit display switching on an enum as the struct version, if used to preference things, doesn't require knowledge of what unit is set.

Addresses Issue https://github.com/MrLotU/SwiftPrometheus/issues/30 but doesn't support both 1.* and 2.* releases, only 2.*

### Description
Switches import of TimeUnit 
XCTest units

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -13,7 +13,7 @@ let package = Package(
             targets: ["PrometheusExample"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-metrics.git", from: "1.2.0"),
+		.package(url: "https://github.com/apple/swift-metrics.git", from: "2.0.0"),
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.0.0"),
     ],
     targets: [
--- a/Package.swift
+++ b/Package.swift
@@ -13,7 +13,7 @@ let package = Package(
             targets: ["PrometheusExample"]),
     ],
     dependencies: [
-		.package(url: "https://github.com/apple/swift-metrics.git", from: "2.0.0"),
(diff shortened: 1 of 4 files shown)


Package.swift
```diff
@@ -13,7 +13,7 @@ let package = Package(
             targets: ["PrometheusExample"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-metrics.git", from: "1.2.0"),
+		.package(url: "https://github.com/apple/swift-metrics.git", from: "2.0.0"),
```
MrLotU: Nitpick, but could you format these indents? 😄

Tests/SwiftPrometheusTests/SummaryTests.swift
```diff
@@ -76,7 +76,8 @@ final class SummaryTests: XCTestCase {
         
         summary.recordSeconds(1)
         summary.recordMilliseconds(2 * 1_000)
-        summary.recordNanoseconds(4 * 1_000_000_000)
+        summary.recordMicroseconds(3 * 1_000_000)
+		summary.recordNanoseconds(4 * 1_000_000_000)
```
MrLotU: Same for this indent

PR #32, opened 2020-04-21
Add LabelSanitizers

### LINKED ISSUE #29: Configurable label separator char replacements
### Feature Request

Offer some configuration to replace characters in labels, i.e. when people in their libraries have `Counter("things.specificThing.thingsToCound")` that's an invalid label for prometheus. So we should allow some `labelSubstitutions = [".": "_"]` or perhaps some other way?

Libraries which offer metrics via Swift Metrics generally "should not know" details about formats of labels that any specific backends require. This is mostly about the use of `.` (prom crashes on those, other backends like those) as separator vs using `_` (prom likes those, other backends dislike those).

#### Alternatives or Workarounds

- No go alternative: force all possible libraries to avoid `.` 😉

MrLotU:
Fixes #29 

### Checklist


Sources/Prometheus/Prometheus.swift
```diff
@@ -16,10 +16,15 @@ public class PrometheusClient {
     /// Lock used for thread safety
     private let lock: Lock
     
+    /// Sanitizers used to clean up label values provided through
+    /// swift-metrics.
+    public let sanitizers: [LabelSanitizer]
+    
     /// Create a PrometheusClient instance
-    public init() {
+    public init(_ sanitizers: [LabelSanitizer] = [PrometheusLabelSanitizer()]) {
```
ktoso, line 24: Are we sure we need multiple sanitizers really?

MrLotU: Good point probably. Will update to just one

ktoso: Cool 👍  (Minor nitpick, perhaps worth keeping the argument label?)

Issue #33, opened 2020-05-15
Add import of UIKit to use the library on an iOS client


matiasvillaverde:
I tried to use the library from an iOS app, and everything is compatible except that the compiler gets a bit confused of where to use`CGFloat`. 

While trying to compile we get this error
` Use of undeclared type 'CGFloat' `

This gets fixed by adding a compiler flag to know if we need to import `CoreGraphics` on iOS.

### Steps to reproduce

1. Import Prometheus library to an iOS App
2. Try to compile the app
3. Error will prevent compiling

### Expected behavior

Just be able to compile on iOS 

* OS version: iOS 12
* Swift version: 5.1
* Serverside Swift Framework: -
* Framework version: 1.0.0-alpha

MrLotU:
Fixed per #34

MrLotU:
Now available in https://github.com/MrLotU/SwiftPrometheus/releases/tag/1.0.0-alpha.7

PR #34, opened 2020-05-15
Prevent the compiler finding the wrong CGFloat

matiasvillaverde:
As open on this [Issue](https://github.com/MrLotU/SwiftPrometheus/issues/33), this small PR allows to use the library on iOS, WatchOS and tvOS. 

### Checklist
- [ ] I've created new tests where needed.

### Motivation and Context

Then we can use the library on iOS, the issue is this [one](https://github.com/MrLotU/SwiftPrometheus/issues/33)

### Description

Adding a compiler flag, to import CoreGraphics will fix the compiler getting crazy of which CGFloat to use.

### DIFF
--- a/Sources/Prometheus/Utils.swift
+++ b/Sources/Prometheus/Utils.swift
@@ -1,5 +1,9 @@
 import Foundation
 
+#if os(iOS) || os(watchOS) || os(tvOS)
+import CoreGraphics
+#endif
+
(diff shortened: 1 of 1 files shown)


MrLotU:
Now available in https://github.com/MrLotU/SwiftPrometheus/releases/tag/1.0.0-alpha.7

PR #35, opened 2020-11-10
Remove CoreGraphics

ktoso:
Hey @MrLotU, thanks in advance for considering this PR. It would help us in some use cases we'd like to use this library 👍 

### Motivation and Context

In some environments it is better to avoid pulling in "dependencies" if we don't really use them. 
There can be situations in which CoreGraphics is not available yet one would still like to use this library.

The conformance is not super crucial to ship with SwiftPrometheus, if someone wants to they can conform GCFloat to `ConvertibleNumberType` manually.

### Description

Remote the not strictly necessary CoreGraphics usage.

### DIFF
--- a/Sources/Prometheus/Utils.swift
+++ b/Sources/Prometheus/Utils.swift
@@ -1,9 +1,5 @@
 import Foundation
 
-#if os(iOS) || os(watchOS) || os(tvOS)
-import CoreGraphics
-#endif
-
 /// Empty labels class
 public struct EmptyLabels: MetricLabels {
     /// Creates empty labels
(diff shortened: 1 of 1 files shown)


MrLotU:
Hi @ktoso. Thanks for the PR. I agree with your point and think this would be a decent change. However, it would result in a breaking change since the `CGFloatValue` that's removed is a public property.

Instead, I would propose to put both the import of `CoreGraphics` and the `CGFloatValue` inside a `canImport` statement to make sure `CoreGraphics` is available. This should then not cause any breaking changes for platforms that _do_ have CoreGraphics, while fixing issues for platforms that don't.

ktoso:
That sounds good, let me confirm that'll definitely work and I'll adjust the PR 👍

ktoso:
I dug into this a bit and it does not seem in the case we're trying to solve.

I really would strongly push for _not_ depending on not-server stuff in libraries like this which are really mostly going to be used in servers. The small inconvenience for someone to manually conform a `CG...` type if they really really need a metric using those types is relatively small, compared to blocking some use cases from using the library.

Since the lib is `-alpha` removals are all fair game still. Think we can get this in?

MrLotU:
For sure agreed there. Making use of `canImport` would've been a best case scenario here, but should, IMO, not be a blocker to move forward with this PR anyways, since we're still in Alpha land.

Thanks for the PR, I'm going to look into the CI failures today and merge, tag & release after.

MrLotU (approved):
LGTM. Thanks! 👍

ktoso:
Thank you very much for merge and quick release 🙏

PR #37, opened 2020-11-17
Update README.md

jordanebelanger:
The Vapor 4 metrics endpoint example leads to a leaked promise error when running it in debug mode. Consider the following instead.


MrLotU (approved):
LGTM! Thanks 👍

PR #41, opened 2021-03-03
fix Buckets helpers to not throw fatal error

avolokhov:
Hi @MrLotU. Thanks for considering this PR.
In my project I needed to create custom buckets and stumbled upon `Buckets` helpers throwing fatal errors all the time. Here's a fix for this.

### Checklist

### Motivation and Context

Buckets static helpers are always throwing fatal errors.

### Description

`Buckets` static helpers are working now.

### DIFF
--- a/Sources/Prometheus/MetricTypes/Histogram.swift
+++ b/Sources/Prometheus/MetricTypes/Histogram.swift
@@ -42,7 +42,7 @@ public struct Buckets: ExpressibleByArrayLiteral {
         var arr = [Double]()
         var s = start
         for x in 0..<count {
-            arr[x] = s
+            arr.append(s)
             s += width
         }
         return Buckets(arr)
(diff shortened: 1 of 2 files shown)


Sources/Prometheus/MetricTypes/Histogram.swift
```diff
@@ -41,7 +41,7 @@ public struct Buckets: ExpressibleByArrayLiteral {
         assert(count >= 1, "Bucket.linear needs a count larger than 1")
         var arr = [Double]()
```
weissi, line 42: random drive-by but we could make all this code

```
let arr = (1...count).map { Double($0) * Double(start) }
```

which would make it slightly more efficient and just one line.

Feel free to ignore.

avolokhov: Sure, why not.

MrLotU (approved):

PR #42, opened 2021-03-03
make labels classes public to allow users to create their own prometh…

avolokhov:
…eus classes

Hi, @MrLotU. Thanks for considering this PR. When adopting SwiftPrometheus in my project, I had to use custom buckets. Current `swift-metrics` don't allow setting buckets directly, so I had to use `Prometheus.createHistogram` directly. Unfortunately, default labels implementation for `createHistogram` is `EmptyHistogramLabels`, and I can't pass `DimensionalHistogramLables` explicitly because they're private to the library. I think making them public will result in better flexibility and allow adopters to enable more use cases and simplify building their own logic on top current SwiftPrometheus.

DimensionLabels, DimensionHistogramLabels, DimensionSummaryLabels are made public

### Checklist
- [n/a] I've created new tests where needed.

### Motivation and Context
swift-metrics api does not allow setting buckets explicitly. This forces adopter either use SwiftPrometheus classes directly or only use default buckets. When default buckets don't fit and direct Prometheus classes are used, using direct Prometheus classes for other types of metrics can be logical and may lead to a more standardised readable code.

### Description
Private labels classes forces library adopter to copy/paste labels classes when they want to use Histograms with SwiftPrometheus metrics types directly. Having them public may simplify the adoption.

### DIFF
--- a/Sources/Prometheus/PrometheusMetrics.swift
+++ b/Sources/Prometheus/PrometheusMetrics.swift
@@ -236,25 +236,25 @@ private struct StringCodingKey: CodingKey {
 
 
 /// Helper for dimensions
-private struct DimensionLabels: MetricLabels {
+public struct DimensionLabels: MetricLabels {
     let dimensions: [(String, String)]
     
-    init() {
+    public init() {
         self.dimensions = []
     }
     
-    init(_ dimensions: [(String, String)]) {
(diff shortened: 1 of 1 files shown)


MrLotU (approved):

PR #43, opened 2021-03-03
New line in reporter

avolokhov:
Hi, @MrLotU. Thanks for considering this PR. 
This PR is to make collection to String and to ByteBuffer identical. I'm not 100% sure, but I think new line in the end is expected (and also my scraping tool complained when it saw it lacking). 

### Checklist
- [n/a] I've updated the documentation if necessary.

### Motivation and Context
In its protocol Prometheus expects new line at the end of string. New line was added when collecting into buffer but not into string. This PR will make collection results identical when collecting into string and into byte buffer.

### Description
Collection into string is now consistent with collection into buffer.
Created a test that checks equality on collected to string and collected to buffer output.


MrLotU (approved):

Issue #44, opened 2021-04-06
Support a histogram backed timer.


avolokhov:
#### Motivation Behind Feature
Hi, @MrLotU. 
When adapting prometheus data types to swift-metrics api, there's multiple ways to back API types with prometheus types. In particular, both `Summary` and `Histogram` can serve as a backend for `Timer`. In their docs, prometheus compares `Summary` and `Histogram` implementations: https://prometheus.io/docs/practices/histograms/#quantiles.
One major difference there is that `Summary` backed values are not aggregatable. For server side application it's common to have multiple replicas of your backend. This makes `Summary` based timer a show stopper.

#### Feature Description
For the reference (it contradicts my point, but I still have to mention it
One solution I see is to make `MetricsFactory` conformance a wrapper and not an extension to `PrometheusClient`. This will make the code much more flexible, allow users to pick/write their own `MetricsFactory` conformance and potentially help us with more elaborate Prometheus -> swift-metrics adapters (see my other issue https://github.com/MrLotU/SwiftPrometheus/issues/36; being able to store `baseLabels` separately from `PrometheusClient` will simplify the solution).

#### Alternatives or Workarounds
It contradicts my point, but I have to mention that [java dropwizard adapter](https://github.com/prometheus/client_java/blob/master/simpleclient_dropwizard/src/main/java/io/prometheus/client/dropwizard/DropwizardExports.java#L144) uses `Summary` as a backed for `Timer`. However, it still forces an anti pattern 
```
avg(http_request_duration_seconds{quantile="0.95"}) // BAD!
```
Prometheus mentions on their docs. 
Python prometheus-flask-exporter [uses](https://pypi.org/project/prometheus-flask-exporter/) Histogram as backing type for timer.

In my local fork I wrap PrometheusClient with a custom metrics factory that does this:
```swift
    private func makeHistogramBackedTimer(label: String, dimensions: [(String, String)]) -> TimerHandler {
        let createHandler = { (histogram: PromHistogram) -> TimerHandler in
            HistogramBackedTimer(histogram: histogram, dimensions: dimensions)
        }
        if let histogram: PromHistogram<Int64, DimensionHistogramLabels> = prometheus.getMetricInstance(with: label, andType: .histogram) {
            return createHandler(histogram)
        }
        return createHandler(prometheus.createHistogram(forType: Int64.self, named: label, buckets: defaultExponentialBuckets, labels: DimensionHistogramLabels.self))
    }

[...]

PR #46, opened 2021-04-19
Histogram backed timer

avolokhov:
Hey @MrLotU,

Thanks for considering this PR. In my project I'm running multiple replicas of my backend service. For a proper monitoring I needed aggregatable timer metrics to get cumulative percentiles across the whole fleet. Here's the solution I came up with. Since the project is in `alpha`, I decided to start with breaking a few APIs in order to keep the code simple. Please let me know if you think we should approach this differently.

Fixes https://github.com/MrLotU/SwiftPrometheus/issues/44

### Checklist

### Motivation and Context
Server side applications are never run in as single instance. Summary backed timer doesn't allow aggregation and therefore doesn't allow to render accumulated latency percentiles across multiple instances. Histogram backed timer provides a solution with it.
Timer implementation is yet another fact that only makes sense for bridging `Prometheus` to `swift-metrics` (in addition to `Labels` that are init parameters in `swift-metrics` world and call parameters in `Prometheus` world, and therefore `LabelSanitizer`). It feels natural to separate non-trivial `swift-metrics` bridge from Prometheus core, keeping either of them simpler and more testable. 

### Description
- MetricsFactory conformance is now a wrapper around PrometheusClient. This allows us to be much more flexible and configurable when bridging `swift-metrics` api to Prometheus backend (i.e. picking a Timer implementation, setting default quantiles/buckets, in future potentially setting base labels).
- Introduced configuration object.
- Introduced `Histogram` backed `Timer` implementation.

### DIFF
--- a/Sources/Prometheus/PrometheusMetrics.swift
+++ b/Sources/Prometheus/PrometheusMetrics.swift
@@ -64,15 +64,9 @@ private class MetricsHistogram: RecorderHandler {
     }
 }
 
-/// This is a `swift-metrics` timer backed by a Prometheus' `PromHistogram` implementation.
-/// This is superior to `Summary` backed timer as `Summary` emits a set of quantiles, which will be impossible to correctly aggregate when one wants to render a percentile for a set of multiple instances.
-/// `Histogram` aggregation is possible with some fancy server-side math.
 class MetricsHistogramTimer: TimerHandler {
     let histogram: PromHistogram<Int64, DimensionHistogramLabels>
     let labels: DimensionHistogramLabels?
-    // this class is a lightweight wrapper around heavy prometheus metric type. This class is not cached and each time
-    // created anew. This allows us to use variable timeUnit without locking.
-    var timeUnit: TimeUnit?
 
     init(histogram: PromHistogram<Int64, DimensionHistogramLabels>, dimensions: [(String, String)]) {
         self.histogram = histogram
@@ -83,15 +77,8 @@ class MetricsHistogramTimer: TimerHandler {
         }
     }
 
-    // this is questionable as display unit here affects how the data is stored, and not how it's observed.
-    // should we delete it and tell preferDisplayUnit is not supported?
-    func preferDisplayUnit(_ unit: TimeUnit) {
-        self.timeUnit = unit
-    }
-
     func recordNanoseconds(_ duration: Int64) {
-        // histogram can't be configured with timeUnits, so we have to modify incoming data
-        histogram.observe(duration / Int64(timeUnit?.scaleFromNanoseconds ?? 1), labels)
+        return histogram.observe(duration, labels)
     }
 }
 
@@ -113,7 +100,7 @@ private class MetricsSummary: TimerHandler {
     }
     
     func recordNanoseconds(_ duration: Int64) {
-        summary.observe(duration, labels)
+        return summary.observe(duration, labels)
     }
 }
 
--- a/Sources/Prometheus/PrometheusMetrics.swift
+++ b/Sources/Prometheus/PrometheusMetrics.swift
@@ -64,6 +64,37 @@ private class MetricsHistogram: RecorderHandler {
     }
 }
 
+/// This is a `swift-metrics` timer backed by a Prometheus' `PromHistogram` implementation.
+/// This is superior to `Summary` backed timer as `Summary` emits a set of quantiles, which will be impossible to correctly aggregate when one wants to render a percentile for a set of multiple instances.
+/// `Histogram` aggregation is possible with some fancy server-side math.
+class MetricsHistogramTimer: TimerHandler {
+    let histogram: PromHistogram<Int64, DimensionHistogramLabels>
+    let labels: DimensionHistogramLabels?
+    // this class is a lightweight wrapper around heavy prometheus metric type. This class is not cached and each time
+    // created anew. This allows us to use variable timeUnit without locking.
+    var timeUnit: TimeUnit?
+
+    init(histogram: PromHistogram<Int64, DimensionHistogramLabels>, dimensions: [(String, String)]) {
+        self.histogram = histogram
+        if !dimensions.isEmpty {
+            self.labels = DimensionHistogramLabels(dimensions)
+        } else {
+            self.labels = nil
+        }
+    }
+
+    // this is questionable as display unit here affects how the data is stored, and not how it's observed.
+    // should we delete it and tell preferDisplayUnit is not supported?
+    func preferDisplayUnit(_ unit: TimeUnit) {
+        self.timeUnit = unit
+    }
+
+    func recordNanoseconds(_ duration: Int64) {
+        // histogram can't be configured with timeUnits, so we have to modify incoming data
+        histogram.observe(duration / Int64(timeUnit?.scaleFromNanoseconds ?? 1), labels)
+    }
+}
+
 private class MetricsSummary: TimerHandler {
     let summary: PromSummary<Int64, DimensionSummaryLabels>
     let labels: DimensionSummaryLabels?
@@ -94,7 +125,7 @@ private class MetricsSummary: TimerHandler {
 ///     let sanitizer: LabelSanitizer = ...
 ///     let prometheusLabel = sanitizer.sanitize(nonPrometheusLabel)
 ///
-/// By default `PrometheusLabelSanitizer` is used by `PrometheusClient`
+/// By default `PrometheusLabelSanitizer` is used by `PrometheusMetricsFactory`
 public protocol LabelSanitizer {
     /// Sanitize the passed in label to a Prometheus accepted value.
     ///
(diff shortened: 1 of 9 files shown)


Sources/Prometheus/PrometheusMetrics.swift
```diff
@@ -121,73 +152,119 @@ public struct PrometheusLabelSanitizer: LabelSanitizer {
...
+public struct PrometheusMetricsFactory: MetricsFactory {
+
+    /// Prometheus client to bridge swift-metrics API to.
+    private let prometheusClient: PrometheusClient
+
+    /// Bridge configuration.
+    private let configuration: PrometheusMetricsFactory.Configuration
+
+    public init(prometheusClient: PrometheusClient,
+                configuration: PrometheusMetricsFactory.Configuration = PrometheusMetricsFactory.Configuration()) {
```
fabianfett, line 167:
```suggestion
                configuration: PrometheusMetricsFactory.Configuration = .init()) {
```

Sources/Prometheus/PrometheusMetrics.swift
```diff
@@ -121,73 +152,119 @@ public struct PrometheusLabelSanitizer: LabelSanitizer {
...
+/// A bridge between PrometheusClient and swift-metrics. Prometheus types don't map perfectly on swift-metrics API,
+/// which makes bridge implementation non trivial. This class defines how exactly swift-metrics types should be backed
+/// with Prometheus types, e.g. how to sanitize labels, what buckets/quantiles to use for recorder/timer, etc.
+public struct PrometheusMetricsFactory: MetricsFactory {
+
+    /// Prometheus client to bridge swift-metrics API to.
+    private let prometheusClient: PrometheusClient
+
+    /// Bridge configuration.
+    private let configuration: PrometheusMetricsFactory.Configuration
```
fabianfett, line 164:
```suggestion
    private let configuration: Configuration
```

Sources/Prometheus/PrometheusMetrics.swift
fabianfett, line 167:
```suggestion
    public init(client: PrometheusClient,
                configuration: Configuration = .init()) {
```

Sources/Prometheus/PrometheusMetrics.swift
```diff
@@ -121,73 +152,119 @@ public struct PrometheusLabelSanitizer: LabelSanitizer {
...
 }
 
-extension PrometheusClient: MetricsFactory {
+/// A bridge between PrometheusClient and swift-metrics. Prometheus types don't map perfectly on swift-metrics API,
+/// which makes bridge implementation non trivial. This class defines how exactly swift-metrics types should be backed
+/// with Prometheus types, e.g. how to sanitize labels, what buckets/quantiles to use for recorder/timer, etc.
+public struct PrometheusMetricsFactory: MetricsFactory {
+
+    /// Prometheus client to bridge swift-metrics API to.
+    private let prometheusClient: PrometheusClient
```
fabianfett, line 161:
```suggestion
    private let client: PrometheusClient
```

fabianfett (commented):
Maybe we can make the API a little more dense?

grennis:
+1 for this.

I tried the branch and it seemed to work great. The only problem I saw was that in Vapor using:

            try MetricsSystem.prometheus().collect(into: promise)

gave me `Fatal error: leaking promise`. But I worked around by using the client I had created:

            client.collect(into: promise)

And this worked.

In any case would love to see this get merged, as providing a histogram backed timer is really a necessity.

Sources/Prometheus/PrometheusMetrics.swift
```diff
@@ -64,6 +64,37 @@ private class MetricsHistogram: RecorderHandler {
     }
 }
 
+/// This is a `swift-metrics` timer backed by a Prometheus' `PromHistogram` implementation.
+/// This is superior to `Summary` backed timer as `Summary` emits a set of quantiles, which will be impossible to correctly aggregate when one wants to render a percentile for a set of multiple instances.
+/// `Histogram` aggregation is possible with some fancy server-side math.
```
MrLotU, line 69:
```suggestion
```

While in general I think it's a good thing to add doc blocks to classes & structs, seeing this is an internal class going along with a handful of other ones without docs, I think we should omit them here too.

Sources/Prometheus/PrometheusMetrics.swift
```diff
@@ -64,6 +64,37 @@ private class MetricsHistogram: RecorderHandler {
...
+class MetricsHistogramTimer: TimerHandler {
+    let histogram: PromHistogram<Int64, DimensionHistogramLabels>
+    let labels: DimensionHistogramLabels?
+    // this class is a lightweight wrapper around heavy prometheus metric type. This class is not cached and each time
+    // created anew. This allows us to use variable timeUnit without locking.
```
MrLotU, line 74:
```suggestion
```

Sources/Prometheus/PrometheusMetrics.swift
```diff
@@ -64,6 +64,37 @@ private class MetricsHistogram: RecorderHandler {
...
+        } else {
+            self.labels = nil
+        }
+    }
+
+    // this is questionable as display unit here affects how the data is stored, and not how it's observed.
+    // should we delete it and tell preferDisplayUnit is not supported?
+    func preferDisplayUnit(_ unit: TimeUnit) {
+        self.timeUnit = unit
+    }
```
MrLotU, line 90: Instead of somewhat supporting `displayUnit`, instead `Histogram` should be updated to take a `displayUnit` into account similar to `Summary`.

Sources/Prometheus/PrometheusMetrics.swift
```diff
@@ -64,6 +64,37 @@ private class MetricsHistogram: RecorderHandler {
...
+
+    func recordNanoseconds(_ duration: Int64) {
+        // histogram can't be configured with timeUnits, so we have to modify incoming data
+        histogram.observe(duration / Int64(timeUnit?.scaleFromNanoseconds ?? 1), labels)
```
MrLotU, line 94: See above comment

Sources/Prometheus/PrometheusMetricsConfiguration.swift
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        var _wrapped: _Wrapped
+
+        private init(_ wrapped: _Wrapped) {
+            self._wrapped = wrapped
+        }
+
+        public static func summary(defaultQuantiles: [Double] = Prometheus.defaultQuantiles) -> TimerImplementation {
+            TimerImplementation(.summary(defaultQuantiles: defaultQuantiles))
```
MrLotU, line 16: This will need a `return` statement for Swift 5
```suggestion
            return TimerImplementation(.summary(defaultQuantiles: defaultQuantiles))
```

Sources/Prometheus/PrometheusMetricsConfiguration.swift
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        public static func histogram(defaultBuckets: Buckets = Buckets.defaultBuckets) -> TimerImplementation {
+            TimerImplementation(.histogram(defaultBuckets: defaultBuckets))
```
MrLotU, line 20: This will need a `return` statement for Swift 5
```suggestion
            return TimerImplementation(.histogram(defaultBuckets: defaultBuckets))
```

Sources/Prometheus/PrometheusMetricsConfiguration.swift
```diff
@@ -0,0 +1,45 @@
...
+
+        /// This parameter will define what implementation will be used for bridging `swift-metrics` to Prometheus types.
+        public var timerImplementation: PrometheusMetricsFactory.TimerImplementation
+
+        /// Default buckets for `Recorder` with aggregation.
+        public var defaultHistogramBuckets: Buckets
+
+        public init(labelSanitizer: LabelSanitizer = PrometheusLabelSanitizer(),
+                    timerImplementation: PrometheusMetricsFactory.TimerImplementation = .summary(),
+                    defaultHistogramBuckets: Buckets = .defaultBuckets) {
```
MrLotU, line 39: Would suggest renaming this to `defaultRecorderBuckets` to indicate these are for `Recorder` instances, _not_ for `Timer`s

MrLotU (changes requested):
Overall a great patch. Thanks a lot!

I left a few style & nit comments, as well as the request to add `timeUnit` support to `Histogram` in the same way `Summary` supports this.

Other than that, LGTM!

MrLotU:
> +1 for this.
> 
> I tried the branch and it seemed to work great. The only problem I saw was that in Vapor using:
> 
> ```
>         try MetricsSystem.prometheus().collect(into: promise)
> ```
> 
> gave me `Fatal error: leaking promise`. But I worked around by using the client I had created:
> 
> ```
>         client.collect(into: promise)
> ```
> 
> And this worked.
> 
> In any case would love to see this get merged, as providing a histogram backed timer is really a necessity.

@grennis The issue you mention about leaking promises seems unrelated to this PR, but should not be happening for sure. Would you mind filing a bug-report with some extra information like system setup, full error message and reproduction steps? You can do so [here](https://github.com/MrLotU/SwiftPrometheus/issues/new/choose)

Thanks!

Sources/Prometheus/PrometheusMetrics.swift
avolokhov: This would be much more difficult as `Summary` produces aggregates on demand, while `Histogram` precomputes values when metric is reported. So changing timeUnit in `Summary` will allow you to produce different time units on the fly, while changing it in `Histogram` will force you to modify data during the reporting. That's why I didn't do it in a first place: there's simply no way to copy and paste `Summary` approach into `Histogram`.

MrLotU: Ah, yes. That's my bad. In that case I would remove the functionality in its entirety. As per [swift-metrics](https://github.com/apple/swift-metrics/blob/main/Sources/CoreMetrics/Metrics.swift#L555-L559):

```
    /// Set the preferred display unit for this TimerHandler.
    ///
    /// - parameters:
    ///     - unit: A hint to the backend responsible for presenting the data of the preferred display unit. This is not guaranteed to be supported by all backends.
```

MrLotU (approved):
Looks great. Thanks for the PR 😄

MrLotU:
Released in [1.0.0-alpha.10](https://github.com/MrLotU/SwiftPrometheus/releases/tag/1.0.0-alpha.10)

PR #50, opened 2021-05-24
Perf improvements

glbrntt:
Motivation:

Metrics collecting systems should be efficient enough so that their
impact on the system being instrumented is negligible. There are 
a few parts of SwiftPrometheus which are a bottleneck when producing
a large number of metrics.

Modifications:

- Store `PromMetric`s in a dictionary keyed by label in
  `PrometheusClient`; this gives fast lookup when checking if a metric
  already exists
- Remove the `metricTypeMap`, we can recover the same information from
  `PromMetric`
- Remove calls to `getMetricInstrumet` in `PrometheusMetricsFactory`
  these were redundant as the same check is done in `createCounter` etc.
- In each `createCounter` (etc.) call we now hold the lock for the
  duration of the call to avoid races between checking the cache and
  creating and storing a new metric.
- The `PrometheusLabelSanitizer` now checks whether input needs
  santizing instead of santizing all input
- Sanitizing is done in one step, mapping each utf8 code point to a
  sanitized code point rather than lowercasing and then checking the
  validity of each character.

Result:

- Sanitizing pre-sanitized labels is ~100x faster
- Sanitizing non-sanitized label is ~20x faster
- Incrementing 10 counters is ~20x faster
- Incrementing 100 counters is ~40x faster
- Incrementing 1000 counters is ~250x faster
- (Similar results for other metrics.)

### Checklist

### DIFF
--- a/Sources/Prometheus/Prometheus.swift
+++ b/Sources/Prometheus/Prometheus.swift
@@ -7,19 +7,14 @@ import NIO
 public class PrometheusClient {
 
     /// Metrics tracked by this Prometheus instance
-    private var metrics: [PromMetric]
-    
-    /// To keep track of the type of a metric since  it can not change
-    /// througout the lifetime of the program
-    private var metricTypeMap: [String: PromMetricType]
+    private var metrics: [String: PromMetric]
     
     /// Lock used for thread safety
     private let lock: Lock
     
     /// Create a PrometheusClient instance
     public init() {
-        self.metrics = []
-        self.metricTypeMap = [:]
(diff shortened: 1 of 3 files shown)


MrLotU (approved):
Thanks a lot for the PR @glbrntt ! Awesome work.

Will get this merged & tagged 👍🏻

MrLotU:
Released in [Alpha 11](https://github.com/MrLotU/SwiftPrometheus/releases/tag/1.0.0-alpha.11)

PR #57, opened 2021-08-02
Introduce PromSummary.capacity to improve performance

### LINKED ISSUE #56: Summary calculation time grows with time (potentially limitless)
The number of `PromSummary.values` is not limited in any way. They are sorted on every`collect()` call and the calculation time grows potentially limitless.

wacumov:
### Motivation and Context
Fixes #56.

### Description
The default buffer size for summary observations in the Go Prometheus client is [500](https://github.com/prometheus/client_golang/blob/20eef740428e9c218cf8637d9b4f4d7e3f5d4d4f/prometheus/summary.go#L80). So I think it's a good choice for the default value for `PromSummary.capacity` in SwiftPrometheus too.

### DIFF
--- a/Sources/Prometheus/MetricTypes/Summary.swift
+++ b/Sources/Prometheus/MetricTypes/Summary.swift
@@ -44,7 +44,10 @@ public class PromSummary<NumType: DoubleRepresentable, Labels: SummaryLabels>: P
     
     /// Values in this Summary
     private var values: [NumType] = []
-    
+
+    /// Number of last values used to calculate quantiles
+    internal let capacity: Int
+
     /// Quantiles used by this Summary
     internal let quantiles: [Double]
     
@@ -60,9 +63,10 @@ public class PromSummary<NumType: DoubleRepresentable, Labels: SummaryLabels>: P
     ///     - name: Name of the Summary
     ///     - help: Help text of the Summary
     ///     - labels: Labels for the Summary
+    ///     - capacity: Number of last values used to calculate quantiles
     ///     - quantiles: Quantiles to use for the Summary
     ///     - p: Prometheus instance creating this Summary
-    internal init(_ name: String, _ help: String? = nil, _ labels: Labels = Labels(), _ quantiles: [Double] = Prometheus.defaultQuantiles, _ p: PrometheusClient) {
+    internal init(_ name: String, _ help: String? = nil, _ labels: Labels = Labels(), _ capacity: Int = Prometheus.defaultSummaryCapacity, _ quantiles: [Double] = Prometheus.defaultQuantiles, _ p: PrometheusClient) {
         self.name = name
         self.help = help
         
@@ -74,6 +78,8 @@ public class PromSummary<NumType: DoubleRepresentable, Labels: SummaryLabels>: P
         
         self.count = .init("\(self.name)_count", nil, 0, p)
         
+        self.capacity = capacity
+
         self.quantiles = quantiles
         
         self.labels = labels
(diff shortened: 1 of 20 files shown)


MrLotU (commented):
Overall PR LGTM, however would appreciate if you could add a quick test making sure everything works as intended.

Thanks!

Sources/Prometheus/MetricTypes/Summary.swift
```diff
@@ -160,6 +166,9 @@ public class PromSummary<NumType: DoubleRepresentable, Labels: SummaryLabels>: P
             self.count.inc(1)
             self.sum.inc(value)
             self.values.append(value)
+            if self.values.count > self.capacity {
+                self.values.remove(at: 0)
```
rauhul, line 170: This is `O(n)` and will have poor performance. It would be better to replace the array here with a `RingBuffer` or `Deque` etc...

wacumov: Thanks for pointing this out! I thought about it but was not sure how to do it the right way. To add a dependency just for this case? To write an own implementation?

After your comment, I have chosen the first option, but swift-collections package's tools version is 5.3 and SwiftPrometheus' tools version is 5.2, so they are incompatible.
Probably, it's a bad idea to change the tools version of this package because of this case.

So, again, I'm not sure what is the best solution in this case.

MrLotU: Best solution is probably to use NIOs `CircularBuffer` type that's an automatically expanding `RingBuffer`, which also has an `O(1)` `popFirst()`.

When using this, we should initialise the `CircularBuffer` with an initial capacity of `capacity + 1` (or `capacity` and rework the adding to pop before adding) to avoid (re-)allocations and copies.

MrLotU: Agree that both pulling in a dependency, and/or increasing tools version for this would be undesirable 😄

wacumov: Thanks for the advice! I didn't know about NIOs `CircularBuffer`.

Sources/Prometheus/MetricTypes/Summary.swift
```diff
@@ -60,9 +64,10 @@ public class PromSummary<NumType: DoubleRepresentable, Labels: SummaryLabels>: P
     ///     - name: Name of the Summary
     ///     - help: Help text of the Summary
     ///     - labels: Labels for the Summary
+    ///     - capacity: Number of last values used to calculate quantiles
```
MrLotU, line 67:
```suggestion
    ///     - capacity: Number of values to keep for calculating quantiles
```

Sources/Prometheus/MetricTypes/Summary.swift
```diff
@@ -43,8 +44,11 @@ public class PromSummary<NumType: DoubleRepresentable, Labels: SummaryLabels>: P
     private let count: PromCounter<NumType, EmptyLabels>
     
     /// Values in this Summary
-    private var values: [NumType] = []
-    
+    private var values: CircularBuffer<NumType>
+
+    /// Number of last values used to calculate quantiles
```
MrLotU, line 49:
```suggestion
    /// Number of values to keep for calculating quantiles
```

MrLotU (commented):
2 minor nits on doc comments. Overall LGTM!

Sources/Prometheus/MetricTypes/Summary.swift
wacumov: Done in the last commit

Sources/Prometheus/MetricTypes/Summary.swift
wacumov: Done in the last commit

MrLotU (approved):
Thanks for the PR! LGTM 😄

wacumov:
Thank you for the comments!

PR #59, opened 2021-08-23
Finer concurrency in 'MetricType.observe' / 'MetricType.collect'

avolokhov:
Every metric type has a lock that's taken for the whole duration of consuming a metric value, and also for the whole duration of observe. This PR puts as much computations out of critical section as we can afford, which results in a significant performance boost.
With these changes, `Lock.withLock` takes only 25% of `PromHistogram.observe`, of which half is locking/unlocking and the other half is `collection.filter`.
Reducing the scope of locking helps bringing `lock.withLock` to 50% of `PromHistogram.observe`. Changing `subHistograms` from array to map, reduces it further to `25%`

### Checklist

### Motivation and Context
This PR brings concurrency optimisations that result in better emission latencies.

### Description
Every MetricType's lock is now only held during access to corresponding state in a thread safe manner, all computations are moved outside critical section.
subHistograms is now a `Map` instead of array.

### DIFF
--- a/Sources/Prometheus/MetricTypes/Histogram.swift
+++ b/Sources/Prometheus/MetricTypes/Histogram.swift
@@ -83,7 +83,7 @@ extension HistogramLabels {
 public class PromHistogram<NumType: DoubleRepresentable, Labels: HistogramLabels>: PromMetric, PrometheusHandled {
     /// Prometheus instance that created this Histogram
     internal weak var prometheus: PrometheusClient?
-    
+
     /// Name of this Histogram, required
     public let name: String
     /// Help text of this Histogram, optional
@@ -99,7 +99,7 @@ public class PromHistogram<NumType: DoubleRepresentable, Labels: HistogramLabels
     internal let upperBounds: [Double]
     
     /// Labels for this Histogram
-    internal private(set) var labels: Labels
+    internal let labels: Labels
     
     /// Sub Histograms for this Histogram
     fileprivate var subHistograms: [PromHistogram<NumType, Labels>] = []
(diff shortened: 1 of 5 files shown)


Sources/Prometheus/MetricTypes/Histogram.swift
```diff
@@ -222,21 +230,16 @@ public class PromHistogram<NumType: DoubleRepresentable, Labels: HistogramLabels
...
-        if histograms.count > 2 { fatalError("Somehow got 2 histograms with the same data type") }
-        if let histogram = histograms.first {
+    fileprivate func getOrCreateHistogram(with labels: Labels) -> PromHistogram<NumType, Labels> {
+        let histogram = self.subHistograms[labels]
+        if let histogram = histogram, histogram.name == self.name, histogram.help == self.help {
             return histogram
         } else {
-            let newHistogram = PromHistogram<T, U>(histogram.name, histogram.help, labels, Buckets(histogram.upperBounds), self)
-            histogram.subHistograms.append(newHistogram)
+            guard let prometheus = prometheus else { fatalError("Lingering Histogram") }
```
ktoso, line 240: Hmm can this actually ever happen or just annoyance because that weak there that shoulnd't really manifest in reality?

avolokhov: Yes, this is a weak reference thing. I think reference to Prometheus is redundant, I'll try to track down its origins and potentially remove if indeed unnecessary.

ktoso: Ah i see the error just moved around, it was there before.

Sources/Prometheus/MetricTypes/Histogram.swift
```diff
@@ -142,68 +142,76 @@ public class PromHistogram<NumType: DoubleRepresentable, Labels: HistogramLabels
...
-                
-                let labelsString = encodeLabels(subHistogram.labels, ["le"])
-                output.append("\(subHistogram.name)_count\(labelsString) \(acc)")
-                
-                output.append("\(subHistogram.name)_sum\(labelsString) \(subHistogram.sum.get())")
-                
-                subHistogram.labels.le = ""
+        let (buckets, subHistograms, labels) = self.lock.withLock {
+            (self.buckets, self.subHistograms, self.labels)
+        }
```
ktoso, line 147: 👍

ktoso (approved):
Nice this looks very good, thanks for measuring and the PR Anton!

MrLotU (approved):
LGTM. Thanks for the PR!

PR #60, opened 2021-08-23
make metrics collection independed from metrics creation

avolokhov:
This PR allows metrics collection to be concurrent to metrics creation.
SwiftPrometheus' metrics types are already thread-safe, there's no need in holding a global lock while calling `.collect` on every single metric. 

### Checklist

### Motivation and Context
Currently metrics collection is done under a global lock and it's stopping new metrics from being created in parallel. Allowing this to happen in parallel will reduce the impact metrics framework is having on the running system.

### Description
`PrometheusClient.collect` now only holds global lock to get a list on metrics in a thread safe manner. `metric.collect` on individual metric is called without a global lock and relying on corresponding metric type to be thread safe.


ktoso (approved):
Good catch, thanks Anton!

MrLotU (approved):
LGTM! Thanks for the PR

PR #61, opened 2021-08-25
move labels equality check out of critical section in summary/histogram

avolokhov:
This PR moves `Labels` equality check outside of `Lock`.
The rationale behind it is following:
App's metrics list is usually bound and don't grow a lot after some time of the app being operational. The code already has this constraint implicitly by the fact that none of `Prometheus.metrics`, `Histogram.subHistograms`, `Summary.subSummaries`, `Counter.metrics` are bounded, and will eventually cause an OOM if grow indefinitely. From the domain standpoint, unique label names and values are discouraged and don't provide a lot of insight to app's behaviour.
With the above constraint we know that after some time of app being operational, all or most of label key-value pairs are already stored in corresponding `Histogram.subHistograms` / `Summary.subSummaries`. This makes getting a subSummary/subHistogram from the map a target for optimisations: getOrCreate flow becomes highly skewed towards `get` part. Suggested change improves performance of `get` part of `getOrCreate` and increases the cost less frequently used `create` part. On my local machine this results in more than 2x throughput boost for the following test setups:
```swift

    func testConcurrent() throws {
        let prom = PrometheusClient()
        let histogram = prom.createHistogram(forType: Double.self, 
                                             named: "my_histogram",
                                             helpText: "Histogram for testing",
                                             buckets: Buckets.exponential(start: 1, factor: 2, count: 63),
                                             labels: DimensionHistogramLabels.self)
        let elg = MultiThreadedEventLoopGroup(numberOfThreads: 8)
        let semaphore = DispatchSemaphore(value: 2)
        let time = DispatchTime.now()
        elg.next().submit {
            while DispatchTime.now().uptimeNanoseconds - time.uptimeNanoseconds < 10_000_000_000 {
                let labels = DimensionHistogramLabels([("myValue", "1")])
                histogram.observe(1.0, labels)
            }
            semaphore.signal()
        }
        elg.next().submit {
            while DispatchTime.now().uptimeNanoseconds - time.uptimeNanoseconds < 10_000_000_000 {
                let labels = DimensionHistogramLabels([("myValue", "2")])
                histogram.observe(1.0, labels)
            }
            semaphore.signal()
        }
        semaphore.wait()
        try elg.syncShutdownGracefully()
        print(histogram.collect())
        print(DispatchTime.now().uptimeNanoseconds - time.uptimeNanoseconds)
    }
```

### Checklist
- [ + ] I've created new tests where needed.

### Motivation and Context
This PR optimises metric emission, which results in higher throughput and lower impact on running system.

### Description
scope of locking is now reduced to only obtain/modify current value of `PromHistogram.subHistograms` / `PromSummary.subSummaries`
If subHistogram/subSummaries for given labels is not yet added to the corresponding `PromHistogram` / `PromSummary`, the lock will be held twice: to get current value and to re-check + store new value.

- Created tests to access/update Summary/Histogram from multiple threads
- Ran TSAN against the change to check for concurrency issues.

### DIFF
--- a/Sources/Prometheus/MetricTypes/Histogram.swift
+++ b/Sources/Prometheus/MetricTypes/Histogram.swift
@@ -6,12 +6,12 @@ import Dispatch
 /// See https://prometheus.io/docs/concepts/metric_types/#Histogram
 public struct Buckets: ExpressibleByArrayLiteral {
     public typealias ArrayLiteralElement = Double
-    
+
     public init(arrayLiteral elements: Double...) {
         self.init(elements)
     }
-    
-    fileprivate init (_ r: [Double]) {
+
+    fileprivate init(_ r: [Double]) {
         if r.isEmpty {
             self = Buckets.defaultBuckets
             return
@@ -24,13 +24,13 @@ public struct Buckets: ExpressibleByArrayLiteral {
         assert(Array(Set(r)).sorted(by: <) == r.sorted(by: <), "Buckets contain duplicate values.")
         self.buckets = r
     }
-    
+
     /// The upper bounds
     public let buckets: [Double]
-        
+
     /// Default buckets used by Histograms
     public static let defaultBuckets: Buckets = [0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10]
-    
+
     /// Create linear buckets used by Histograms
     ///
     /// - Parameters:
@@ -42,7 +42,7 @@ public struct Buckets: ExpressibleByArrayLiteral {
         let arr = (0..<count).map { Double(start) + Double($0) * Double(width) }
         return Buckets(arr)
     }
-    
+
     /// Create exponential buckets used by Histograms
     ///
     ///  - Parameters:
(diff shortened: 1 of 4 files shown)


Sources/Prometheus/MetricTypes/Histogram.swift
```diff
@@ -233,14 +233,23 @@ public class PromHistogram<NumType: DoubleRepresentable, Labels: HistogramLabels
 
     /// Helper for histograms & labels
     fileprivate func getOrCreateHistogram(with labels: Labels) -> PromHistogram<NumType, Labels> {
-        let histogram = self.subHistograms[labels]
-        if let histogram = histogram, histogram.name == self.name, histogram.help == self.help {
-            return histogram
+        let subHistograms = lock.withLock { self.subHistograms }
+        if let histogram = subHistograms[labels] {
+            if histogram.name == self.name, histogram.help == self.help {
```
glbrntt, line 238: If we don't believe this can ever fail should we `assert`/`precondition` instead?

Sources/Prometheus/MetricTypes/Histogram.swift
```diff
@@ -233,14 +233,23 @@ public class PromHistogram<NumType: DoubleRepresentable, Labels: HistogramLabels
...
+            } else {
+                fatalError("Somehow got 2 summaries with the same data type")
+            }
         } else {
-            guard let prometheus = prometheus else { fatalError("Lingering Histogram") }
-            let newHistogram = PromHistogram(self.name, self.help, labels, Buckets(self.upperBounds), prometheus)
-            self.subHistograms[labels] = newHistogram
-            return newHistogram
+            return lock.withLock {
+                if let histogram = subHistograms[labels], histogram.name == self.name, histogram.help == self.help {
```
glbrntt, line 245: Above we `fatalError` if name and help don't match up. Do we want to do the same here?

Sources/Prometheus/MetricTypes/Summary.swift
```diff
@@ -190,22 +192,30 @@ public class PromSummary<NumType: DoubleRepresentable, Labels: SummaryLabels>: P
...
-            guard metric.name == summary.name, metric.help == summary.help, metric.labels == labels else { return false }
-            return true
-        }
-        if summaries.count > 2 { fatalError("Somehow got 2 summaries with the same data type") }
-        if let summary = summaries.first {
-            return summary
+    fileprivate func getOrCreateSummary(withLabels labels: Labels) -> PromSummary<NumType, Labels> {
+        let subSummaries = self.lock.withLock { self.subSummaries }
+        if let summary = subSummaries[labels] {
+            if summary.name == self.name, summary.help == self.help {
```
glbrntt, line 198: Should this be an `assert`/`precondition` as above?

MrLotU (approved):
LGTM! Thanks for the PR 😄

PR #64, opened 2021-09-01
fix display unit

avolokhov:
A small concurrency fix for a bug introduced in https://github.com/MrLotU/SwiftPrometheus/pull/59

### Checklist
- [ + ] I've created new tests where needed.

### Motivation and Context
just a small bug fix for thread safe access to summary `displayUnit`

### Description
I touched display unit in one of the tests so it will be caught by TSAN if it ever breaks again.


ktoso (approved):
Good catch, thanks for following up on it

Issue #38, opened 2020-12-07
dimensions are not escaped


AZm87:
I developed a API (based on vapor 4) that throws errors with additional information
```Swift
enum APIError: Error {
    ...
    case invalidParameter(String),
    ...
}
```

I want to monitor the reasons for a failed response with prometheus so I added the following line:
```Swift
prometheus.makeCounter(label: "response_error", dimensions: [("error", "\(error)")]).increment(by: 1)
```

But this results in the following line in the output:
```
response_error{error="invalidParameter("Invalid value for parameter <code>")"} 28
```
According to the [documentation](https://prometheus.io/docs/concepts/data_model/) the "Label values may contain any Unicode characters." but Prometheus can't read this.

### Steps to reproduce

```Swift
import Prometheus

let prometheus = PrometheusClient()
prometheus.makeCounter(label: "response_error", dimensions: [("error", "a\"b\"c")]).increment(by: 1)
let f : (String) -> () = { print($0) }
prometheus.collect(f)
```

### Actual behavior

the result is `response_error{error="a"b"c"} 1`

### Expected behavior

the result should be escaped to `response_error{error="a'b'c"} 1`. In my first tests `response_error{error="a\"b\"c"} 1` also worked but not sure.

### Environment

* OS version: MacOS Big Sur Version 11.0.1
* Swift version: Apple Swift version 5.3.1 (swiftlang-1200.0.41 clang-1200.0.32.8)
* SwiftPrometheus 1.0.0-alpha

MrLotU:
Thanks for opening this issue, and my apologies for the time it took me to respond!

From my initial research I couldn't find anywhere where Prometheus mentions `"` should be an illegal character in metrics, but it does make sense. I will dig a bit deeper into how this can be solved, and also take a look at what other languages' prometheus clients do in this case.

PR #68, opened 2022-06-17
Sanitize Dimensions

### LINKED ISSUE #65: Add dimension name sanitizer
### Feature Request

Add _dimension name_ sanitizer along already existing `LabelSanitizer`.

Dimension names are limited by Prometheus:

> may contain ASCII letters, numbers, as well as underscores. They must match the regex `[a-zA-Z_][a-zA-Z0-9_]*`
>
> names beginning with `__` are reserved for internal use.

Note: Dimension values may contain any Unicode characters.
Note: The metric name (sanitized by `LabelSanitizer`) follows different rules (`[a-zA-Z_:][a-zA-Z0-9_:]*`).

The reasoning behind this feature is same as in the https://github.com/MrLotU/SwiftPrometheus/issues/29.

0xTim:
Sanitize dimensions to avoid scraping from breaking. Resolves #65 and #38

### Checklist

### Motivation and Context
If dimensions aren't sanitized then Prometheus breaks with errors like

```
error while linting: text format parsing error in line 9: expected '=' after label name, found '-'
```

### Description
 Run dimensions through a sanitizer. Uses a new `DimensionsSanitizer` as uppercase ASCII [is allowed](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels). (It looks like regular labels should allow uppercase as well but I didn't want to change that).

Also split out sanitiser types into their own files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 build/
 .swiftpm/
 .idea
+.vscode/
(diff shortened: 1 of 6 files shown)


MrLotU (approved):
LGTM. Thanks!

PR #75, opened 2022-07-31
Use defaultRecorderBuckets for Histogram

### LINKED ISSUE #74: defaultRecorderBuckets not respected with swift-metrics
### Steps to reproduce

Setup Prometheus and Swift Metrics:

```
        let client = PrometheusClient()
        
        // Setup Factory
        let factory = PrometheusMetricsFactory(client: client, configuration: .init(defaultRecorderBuckets: [
            0.05, 0.1, 0.5, 1, 5, 10, 25, 50, 75, 100
        ]))
        
        MetricsSystem.bootstrap(factory)
        
        // Collect Metrics
        app.get("metrics") { _ async throws -> String in
            return await client.collect()
        }
```

Create a Recorder and record a value:

```
Recorder(label: "x").record(100)
```

View output and see that bucket sizes are the default bucket sizes rather than the ones I've provided in the setup.

### Expected behavior

Bucket sizes should be those which I provided to the factory

### Actual behavior

Bucket sizes are instead the defaulted bucket sizes hard coded into the library

### Environment

* OS version: 12.4
* Swift version: 5.6
* Serverside Swift Framework: Vapor
* Framework version: 4

### Solution

The root cause appears to be that the `makeHistogram` function calls `createHistogram` which has a variable called `buckets: Buckets = .defaultBuckets`. This defaults to defaultBuckets which is incorrect, it should instead be set to `configuration.defaultRecorderBuckets` (which itself defaults to `defaultBuckets` but can be overridden by the user, me, as above).

Sherlouk:
### Checklist
- [ ] I've created new tests where needed.

I've not updated tests (mostly because I entirely made this change on GitHub) but we totally should to prevent regression. Created this branch so I can do a quick validation by pointing my project to this branch.

### Motivation and Context

Fixes #74

### Description

Pass the user-provided default for the buckets to the create histogram function. This value defaults to the same value which the function defaults to - so unless the user has overridden the default buckets value, this change will not impact existing code.

### DIFF
--- a/Sources/Prometheus/PrometheusMetrics.swift
+++ b/Sources/Prometheus/PrometheusMetrics.swift
@@ -204,7 +204,7 @@ public struct PrometheusMetricsFactory: PrometheusWrappedMetricsFactory {
 
     private func makeHistogram(label: String, dimensions: [(String, String)]) -> RecorderHandler {
         let label = configuration.labelSanitizer.sanitize(label)
-        let histogram = client.createHistogram(forType: Double.self, named: label)
+        let histogram = client.createHistogram(forType: Double.self, named: label, buckets: configuration.defaultRecorderBuckets)
         return MetricsHistogram(histogram: histogram, dimensions: dimensions.sanitized())
     }
 


ktoso:
Huh this has become quite stale, sorry about this -- it seems to be resolved in other ways on current main.

PR #78, opened 2022-10-17
Fix Histogram Concurrency Issue

Sam-Amin:
Issue:
In Histogram, some values were being dropped when the histogram is accessed concurrently by multiple threads.

Fix:
In getOrCreateHistogram, the code was copying the dictionary into a local variable, and checking it for the label without a lock. If not there, it would take the lock again to add it. But in line 224, after taking the lock again, it was rechecking the local variable, which of course would not have changed in the meantime. But the original dictionary could have changed, and that's the one we should check the second time.

Improved the concurrency test, as it was missing the bug by only checking the total count and sum without labels.

### DIFF
--- a/Sources/Prometheus/MetricTypes/Histogram.swift
+++ b/Sources/Prometheus/MetricTypes/Histogram.swift
@@ -221,7 +221,7 @@ public class PromHistogram<NumType: DoubleRepresentable>: PromMetric {
             return histogram
         } else {
             return lock.withLock {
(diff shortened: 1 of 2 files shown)


ktoso (approved):

ktoso:
Nice catch, thank you @Sam-Amin !

Sorry you've hit this but very thankful for the PR :)

ktoso:
I'll cut a 1.0.1 with this

PR #82, opened 2023-05-26
fix: use lock for subsummary to get its values

blindspotbounty:
There is a race condition listed in https://github.com/swift-server-community/SwiftPrometheus/issues/79
### Checklist
- [-] I've created new tests where needed.
- [-] I've updated the documentation if necessary.

### Motivation and Context
That solves race-condition and corresponding crash from https://github.com/swift-server-community/SwiftPrometheus/issues/79


ktoso (approved):
Thanks for the fix. This looks correct. 

We'd love to make a bigger cleanup here so such mistakes can't happen, but for now let's fix the issue at hand. Thanks!

ktoso:
Uhm, wrong target branch -- we're developing on `main`, I'll pick it over though 🍒 ⛏️ :)

PR #88, opened 2023-09-08
Remove FUNDING.yml

fabianfett:
Jari (the original maintainer) has left the project. This patch removes the GitHub sponsor option that sends money his way.


ktoso (approved):
Agreed, it's only fair since the project was entirely handed over that we don't do sponsorship over here anymore. (None were active at point of merging this anyway).

Also disabled in repo settings.

PR #89, opened 2023-09-08
Remove XCTestManifest & LinuxMain

fabianfett:
We use `--enable-test-discovery` to run tests, no `XCTestManifest` and `LinuxMain` needed.


ktoso (approved):
Nice to get rid of these

PR #92, opened 2023-09-14
New Version 2.0 API

fabianfett:
### Motivation

Swift Prometheus has been in development since 2018. The previous maintainer (@MrLotU) stepped down about a year ago. Since then @ktoso and @fabianfett have maintained the existing code base. 

In order to improve performance in the Prometheus library, we want to replace the existing implementation with a new one for version 2.0. While we do this we also want to get the general API closer to what the prometheus project recommends in their [Writing client libraries](https://prometheus.io/docs/instrumenting/writing_clientlibs/) guide. This means that we will also start to enforce Prometheus rules. For example having two counter with the same name, but one with labels and one without, is illegal in the Prometheus format. Lastly we want to remove the SwiftNIO dependency that is not needed anymore.

I'm well aware that this is a very large PR that changes lots of moving things.

### Changes

- Remove SwiftNIO dependency. Vendor in NIOLock from `NIOConcurrencyHelpers`
- Replace `PrometheusClient` with `PrometheusCollectorRegistry`
- Remove `Summary` implementation
- Replace `Counter`, `Gauge` and `Histogram` implementations
- Add Sendability annotations
- Require Swift 5.7
- Enforces that label names remain constant for a metric name. 

### Performance difference

In simple testing scenarios:

- Exporting values from 1k counters into the prometheus format: ~10x speedup
- Exporting values from 1k gauges into the prometheus format: ~10x speedup
- Exporting values from 1k histograms into the prometheus format: ~25x speedup

### Differences to [Writing client libraries guide](https://prometheus.io/docs/instrumenting/writing_clientlibs/)

- There is no explicit `Collector` protocol, that users could implement to create their own instruments.
	- Counter, Gauge and Histogram should implement `Collector`
- Currently the `PrometheusCollectorRegistry` does not offer a `register` and `unregister` method for `Collector`s
- There is no API to create a Counter/Gauge/Histogram that is not registered with a `PrometheusCollectorRegistry`
- Collectors can't be registered with multiple different `PrometheusCollectorRegistry`s
- We don't advertise a default `collector`

Do we want to offer this API? With the current patch we could add this API later.


.swiftformat
```diff
@@ -0,0 +1,24 @@
+# file options
```
FranzBusch, line 1: In all of our recent projects, we actually adopted `swift-format` since it has a more stable formatting between versions. Would recommend picking it here as well.

Sources/Prometheus/Counter.swift
```diff
@@ -0,0 +1,90 @@
...
+// See CONTRIBUTORS.txt for the list of SwiftPrometheus project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Atomics
+import CoreMetrics
+
+public final class Counter: Sendable {
```
FranzBusch, line 18: Oh we require that all metric backing handlers are classes? This is a bit sad since this could just be a struct with reference semantics.

Second question, does this need to be public?

Sources/Prometheus/Counter.swift
```diff
@@ -0,0 +1,90 @@
...
+    let name: String
+    let labels: [(String, String)]
+    let prerenderedExport: [UInt8]
+
+    init(name: String, labels: [(String, String)]) {
+        self.name = name
+        self.labels = labels
+
+        var prerendered = [UInt8]()
+        prerendered.reserveCapacity(64)
```
FranzBusch, line 31: Could you comment why we need to reserve 64 here?

Sources/Prometheus/Docs.docc/index.md
```diff
@@ -0,0 +1,19 @@
+# ``Prometheus``
+
+A prometheus client library for Swift.
+
+## Overview
```
FranzBusch, line 5: Can we add installation instructions to the index here as well. Allows us to point people directly at the documentation.

Sources/Prometheus/Counter.swift
```diff
@@ -0,0 +1,90 @@
...
+            prerendered.append(contentsOf: #"} "#.utf8)
+        } else {
+            prerendered.append(UInt8(ascii: " "))
+        }
+
+        self.prerenderedExport = prerendered
+    }
+
+    public func increment() {
+        self.increment(by: Int64(1))
```
FranzBusch, line 45: NIT:
```suggestion
        self.increment(by: 1)
```

Sources/Prometheus/Counter.swift
```diff
@@ -0,0 +1,90 @@
...
+import Atomics
+import CoreMetrics
+
+public final class Counter: Sendable {
+    let intAtomic = ManagedAtomic(Int64(0))
+    let floatAtomic = ManagedAtomic(Double(0).bitPattern)
+
+    let name: String
+    let labels: [(String, String)]
+    let prerenderedExport: [UInt8]
```
FranzBusch, line 24: Can those be `private`?

Sources/Prometheus/Histogram.swift
```diff
@@ -0,0 +1,174 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftPrometheus project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import CoreMetrics
+
+public protocol Bucketable: AdditiveArithmetic, Comparable {
```
FranzBusch, line 17: Does this need to be `public`? I guess in general, do we expect people to use all those types without going through `swift-metrics`?

Sources/Prometheus/Histogram.swift
```diff
@@ -0,0 +1,174 @@
...
+    var bucketRepresentation: String { get }
+}
+
+public typealias TimeHistogram = Histogram<Duration>
+public typealias ValueHistogram = Histogram<Double>
+
+public final class Histogram<Value: Bucketable>: @unchecked Sendable {
+    private let lock = NIOLock()
```
FranzBusch, line 25: Can we please use a `NIOLockedValueBox` here. We are really really encouraging everyone to not just put `@unchcked Sendable` on their types but rather stick everything into the locked value box since that only has condition `Sendable` conformance and we have come across a few places where people used a lock and incorrectly declared `@unchecked Sendable`.

Sources/Prometheus/PrometheusCollectorRegistry.swift
```diff
@@ -0,0 +1,557 @@
...
+import CoreMetrics
+
+/// A promtheus collector registry to create and store collectors.
+///
+/// It creates and stores collectors. Further you can use the ``PrometheusCollectorRegistry/emit(into:)``
+/// method to export the metrics form registered collectors into a Prometheus compatible format.
+///
+/// To use a ``PrometheusCollectorRegistry`` with `swift-metrics` use the ``PrometheusMetricsFactory``.
+public final class PrometheusCollectorRegistry: @unchecked Sendable {
+    // Sendability is enforced through the internal `lock`
```
FranzBusch, line 24: Same here please

Sources/Prometheus/PrometheusMetricsFactory.swift
```diff
@@ -0,0 +1,132 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftPrometheus project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import CoreMetrics
+
+/// A wrapper around ``PrometheusClient`` to implement the `swift-metrics` `MetricsFactory` protocol
```
FranzBusch, line 17: There is no `PrometheusClient` anymore

Sources/Prometheus/PrometheusMetricsFactory.swift
```diff
@@ -0,0 +1,132 @@
...
+public struct PrometheusMetricsFactory: Sendable {
+    /// The underlying ``PrometheusClient`` that is used to generate
```
FranzBusch, line 19: Same here ;)

Sources/Prometheus/PrometheusMetricsFactory.swift
```diff
@@ -0,0 +1,132 @@
...
+    /// The default histogram buckets for a ``ValueHistogram``. If there is no explicit overwrite
+    /// via ``valueHistogramBuckets``, the buckets provided here will be used for any new
+    /// Swift Metrics Summary type.
+    public var defaultValueHistogramBuckets: [Double]
+
+    public var valueHistogramBuckets: [String: [Double]]
+
+    public var labelAndDimensionSanitizer: @Sendable (_ label: String, _ dimensions: [(String, String)]) -> (String, [(String, String)])
+
+    public init(client: PrometheusCollectorRegistry) {
```
FranzBusch, line 37: Can we get some docs on those three things?

FranzBusch: Also do I have to create two things now to bootstrap prometheus? Both the `registry` and the `factory`? Couldn't we just default this parameter here to `.init()`?

FranzBusch (commented):
Really nice! Left some comments inline

Sources/Prometheus/Counter.swift
fabianfett: > Oh we require that all metric backing handlers are classes? This is a bit sad since this could just be a struct with reference semantics.

I eventually want to change swift-metrics to not require AnyObject anymore. This will make NoOps much cheaper.

> Second question, does this need to be public?

Yes. I think users should be able to use the Prometheus lib without going through swift-metrics.

Sources/Prometheus/Docs.docc/index.md
fabianfett: Can we do this in a follow up?

Sources/Prometheus/Histogram.swift
fabianfett: Yes, I think there might be use-cases outside using this as a backend for `swift-metrics`. We can make this non-public but then we must provide two explicit Histogram implementations.

Sources/Prometheus/PrometheusMetricsFactory.swift
fabianfett: I wonder if we should create a new one, or if we should provide a `.default` singleton. I think I'd be more in favor of the `.default`, wdyt?

Sources/Prometheus/Histogram.swift
FranzBusch: Fine by me but a bit of documentation wouldn't hurt on this since it is something that is in addition to the normal `swift-metrics` stuff.

Sources/Prometheus/PrometheusMetricsFactory.swift
FranzBusch: I don't see a big benefit on a default singleton here and we avoided introducing singletons in AHC because we want to discuss this in the SSWG in a broader scope. So I would say we shouldn't start introducing it here. Just `.init()` default seems fine to me.

Sources/Prometheus/Counter.swift
ktoso: Agreed on both, eventually we can do a revamp of metrics lib. It's one of the earliest and we've learned much since.

Yes, types should be public, prom has more detailed semantics than swift metrics so an "app" can definitely use it direcly

Sources/Prometheus/Counter.swift
```diff
@@ -0,0 +1,90 @@
...
+        self.name = name
+        self.labels = labels
+
+        var prerendered = [UInt8]()
+        prerendered.reserveCapacity(64)
+        prerendered.append(contentsOf: name.utf8)
+        if let prerenderedLabels = Self.prerenderLabels(labels) {
+            prerendered.append(UInt8(ascii: "{"))
+            prerendered.append(contentsOf: prerenderedLabels)
+            prerendered.append(contentsOf: #"} "#.utf8)
```
ktoso, line 36: can't use ascii here?

Sources/Prometheus/Counter.swift
```diff
@@ -0,0 +1,90 @@
...
+
+        self.prerenderedExport = prerendered
+    }
+
+    public func increment() {
+        self.increment(by: Int64(1))
+    }
+
+    public func increment(by amount: Int64) {
+        precondition(amount >= 0)
```
ktoso, line 49: Can we give those preconditions good messages always please? It's much nicer user experience when crashes contain message

Sources/Prometheus/Counter.swift
```diff
@@ -0,0 +1,90 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Atomics
+import CoreMetrics
+
+public final class Counter: Sendable {
+    let intAtomic = ManagedAtomic(Int64(0))
+    let floatAtomic = ManagedAtomic(Double(0).bitPattern)
```
ktoso, line 20: nit but it's weird to call a Double property "float"; yeah "floating point" but it's a double -> 
```suggestion
    let doubleAtomic = ManagedAtomic(Double(0).bitPattern)
```

Sources/Prometheus/Counter.swift
ktoso, line 24: That's the key to the perf wins I suppose, good 👍

Sources/Prometheus/Counter.swift
```diff
@@ -0,0 +1,90 @@
...
+        precondition(amount >= 0)
+        self.intAtomic.wrappingIncrement(by: amount, ordering: .relaxed)
+    }
+
+    public func increment(by amount: Double) {
+        precondition(amount >= 0)
+        // We busy loop here until we can update the atomic successfully.
+        // Using relaxed ordering here is sufficient, since the as-if rules guarantess that
+        // the following operations are executed in the order presented here. Every statement
+        // depends on the execution before.
```
ktoso, line 58: It's clear here but thx for comments around busy loops, always good to document those 💯

Sources/Prometheus/Gauge.swift
```diff
@@ -0,0 +1,85 @@
...
+// See CONTRIBUTORS.txt for the list of SwiftPrometheus project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Atomics
+import CoreMetrics
+
+public final class Gauge: Sendable {
```
ktoso, line 18: In follow up please add docs to all those types

Sources/Prometheus/Gauge.swift
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        self.prerenderedExport = prerendered
+    }
+
+    public func set(to value: Double) {
+        self.atomic.store(value.bitPattern, ordering: .relaxed)
+    }
+
+    public func increment(by amount: Double = 1.0) {
```
ktoso, line 47: A bit inconsistent, on purpose? The counter had Int64 variants of APIs as well, no need here?

Sources/Prometheus/Histogram.swift
```diff
@@ -0,0 +1,181 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftPrometheus project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import CoreMetrics
+
+public protocol Bucketable: AdditiveArithmetic, Comparable, Sendable {
```
ktoso, line 17: Docs for all those types

Sources/Prometheus/Histogram.swift
```diff
@@ -0,0 +1,181 @@
...
+    var bucketRepresentation: String { get }
+}
+
+public typealias TimeHistogram = Histogram<Duration>
```
ktoso, line 21: Hmmm.... not really "Time", or is it from prom specs...? If not, then mirror the contained value maybe?

Sources/Prometheus/PrometheusCollectorRegistry.swift
```diff
@@ -0,0 +1,556 @@
...
+
+                store[name] = .valueHistogramWithLabels(labelNames, [LabelsKey(labels): histogram], buckets)
+                return histogram
+            }
+        }
+    }
+
+    // MARK: Destroying Metrics
+
+    public func destroyCounter(_ counter: Counter) {
```
ktoso, line 392: They're pretty clear but please document.

E.g. if one can make a new one after a destroy etc

Sources/Prometheus/PrometheusCollectorRegistry.swift
```diff
@@ -0,0 +1,556 @@
...
+            default:
+                return
+            }
+        }
+    }
+
+    // MARK: Emitting
+
+    public func emit(into buffer: inout [UInt8]) {
+        let metrics = self.box.withLockedValue { $0 }
```
ktoso, line 463: I know this is the NIO type, but tbh for such there should be a get() on it, WDYT?

ktoso (approved):
LGTM overall, minor comments inline.

Andrewangeta:
Looks nice! :shipit:

.swiftformat
fabianfett: Can we change this in a follow up?

Sources/Prometheus/Counter.swift
fabianfett: Leads to `Ambiguous use of 'increment(by:)'`

Sources/Prometheus/Histogram.swift
fabianfett: Renamed to `DurationHistogram`

Sources/Prometheus/Gauge.swift
fabianfett: nope not un porpose

ktoso:
Feel free to merge at will btw 👍

Sources/Prometheus/Counter.swift
FranzBusch: FWIW crashed don't contain precondition messages. Only fatal errors messages are included. I have personally switched over to fatal error exclusively because of that reason

FranzBusch (commented):
Everything looks good but we are missing an implementation of the newly added `Meter` type.

FranzBusch (approved):

PR #98, opened 2023-09-18
Don't require to pass a registry on MetricsFactory initialization

fabianfett:
Do not require users to explicitly pass a `PrometheusCollectorRegistry` on `PrometheusMetricsFactory` creation.


Sources/Prometheus/PrometheusMetricsFactory.swift
```diff
@@ -16,8 +16,16 @@ import CoreMetrics
...
+    private static let _defaultRegistry = PrometheusCollectorRegistry()
+
+    /// The default ``PrometheusCollectorRegistry``, which is used inside the ``PrometheusMetricsFactory``
+    /// if no other is provided in ``init(client:)`` or set via ``PrometheusMetricsFactory/client``
+    public static var defaultRegistry: PrometheusCollectorRegistry {
+        self._defaultRegistry
+    }
+
+    /// The underlying ``PrometheusCollectorRegistry`` that is used to generate the swift-metrics handlers
+    public var registry: PrometheusCollectorRegistry
```
FranzBusch, line 28: Do we want this to be public?

fabianfett: Yes. It should! So that users can access what they got by default.

ktoso: I don't mind I guess :)

ktoso (approved):

ktoso:
conflict but LGTM

PR #101, opened 2023-09-22
Document how to use labels

fabianfett:


FranzBusch (approved):

Sources/Prometheus/Docs.docc/labels.md
```diff
@@ -0,0 +1,84 @@
...
+  counter400.increment()
+default:
+  break
+}
+```
+
+> Important: Please note, that all metrics with the same name, **must** use the same label names. 
+
+Prometheus requires that for the same metric name all labels names must be the same. Swift 
+Prometheus ensures that by crashing, if the label names or the metric type does not match a
```
hamzahrmalik, line 45:
```suggestion
Prometheus enforces this by crashing if the label names or the metric type does not match a 
```

Sources/Prometheus/Docs.docc/labels.md
```diff
@@ -0,0 +1,84 @@
+# Use Labels in Swift Prometheus
+
+Learn about the benefits and gotchas of using Labels in Prometheus.
```
fabianfett, line 3:
```suggestion
Learn how to use Labels in Prometheus, the benefits of doing so, and how to avoid common mistakes.
```

hamzahrmalik: might read nicer
```suggestion
Learn how to use Labels in Prometheus, the benefits of doing so, and how to avoid common mistakes
```

hamzahrmalik (approved):

PR #102, opened 2023-09-25
Rename destroyMetric to unregisterMetric

fabianfett:
Renamed the `PrometheusCollectorRegistry` `destroy` methods to `unregister`. this is more in line with the Prometheus [documentation](https://prometheus.io/docs/instrumenting/writing_clientlibs/#overall-structure):

> CollectorRegistry SHOULD offer `register()`/`unregister()` functions, and a Collector SHOULD be allowed to be registered to multiple CollectorRegistrys.

Further this is more in line with what this function actually does. The metrics aren't move only types, that we can actually destroy in the method.

Also added docc comments.

### DIFF
--- a/Sources/Prometheus/PrometheusCollectorRegistry.swift
+++ b/Sources/Prometheus/PrometheusCollectorRegistry.swift
@@ -386,8 +386,13 @@ public final class PrometheusCollectorRegistry: Sendable {
     }
 
     // MARK: Destroying Metrics
-
-    public func destroyCounter(_ counter: Counter) {
+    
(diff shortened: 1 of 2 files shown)


FranzBusch (commented):
Are you also planning to rename the `make` methods?

fabianfett:
> Are you also planning to rename the `make` methods?

No, I don't plan this. Reason: We might want to add explicit calls to `register` and allow create metric creation without a registry attached from the get go. However we could change the internal make implementation to do exactly that.

My goal with this is to get the API in such a shape, that we can be fully compatible with the prometheus requirements without breaking API.

FranzBusch (approved):

ktoso (approved):
Okey so both names exist then, that's ok

PR #103, opened 2023-09-26
Remove generic argument from Histogram

fabianfett:
In #92, we added a `Histogram` that is generic over the observed type. This means that each `Histogram` was also carrying a generic argument. To make it easier to deal with different `Histogram` types, we added two typealiases `DurationHistogram` and `ValueHistogram`.

The motivation behind using a generic argument was to preserve as much detailed information as possible. `Duration` can be more precise than `Double`. However in the grand scheme of things we now believe that this is overkill for a `Histogram` and a simpler to use `Histogram` type is more valuable than the lost precision. Thus this patch removes the generic argument from `Histogram`.

### DIFF
--- a/Sources/Prometheus/Docs.docc/swift-metrics.md
+++ b/Sources/Prometheus/Docs.docc/swift-metrics.md
@@ -111,28 +111,19 @@ generated in a third party library.
 
 #### Default buckets
 
-Swift Metric ``Timer``s are backed by a Prometheus ``DurationHistogram`` and Swift Metric 
-``Recorder``s that aggregate are backed by a Prometheus ``ValueHistogram``. As a user, you can 
+Swift Metric ``Timer``s are backed by a Prometheus ``Histogram`` and Swift Metric 
+``Recorder``s that aggregate are also backed by a Prometheus ``Histogram``. As a user, you can 
 specify which buckets shall be used within the backing ``Histogram``s.
 
 ```swift
 var factory = PrometheusMetricsFactory()
 
-factory.defaultDurationHistogramBuckets = [
-  .milliseconds(5),
-  .milliseconds(10),
-  .milliseconds(25),
-  .milliseconds(50),
-  .milliseconds(100),
+factory.defaultTimerHistogramBuckets = [
+  0.005, 0.01, 0.025, 0.05, 0.1
 ]
 
-factory.defaultValueHistogramBuckets = [
-  5,
-  10,
-  25,
-  50,
-  100,
-  250,
+factory.defaultRecorderHistogramBuckets = [
+  5, 10, 25, 50, 100, 250,
 ]
 MetricSystem.bootstrap(factory)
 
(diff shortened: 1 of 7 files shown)


FranzBusch (approved):

ktoso (changes requested):
I don't think this is the right direction. 

I know doubles are "good enough" for time but but the API exposed as doubles always is really ugly. 

We also now have an `observe(Duration)` on ANY histogram, even if it is actually e.g. sizes of things, which is rather ugly. The split was useful IMHO and I'm not sure why we should pull back on it - the PR doesn't really provide sufficient argument tbh.

If we just want to simplify implementation I'm sure we could just implement an underlying type over Double and expose two APIs for Double and specialized for Duration if that's the primary reason driving this change

Sources/Prometheus/Histogram.swift
```diff
@@ -67,20 +55,29 @@ public final class Histogram<Value: Bucketable>: Sendable {
             state.count += 1
         }
     }
+
+    public func observe(_ value: Duration) {
```
ktoso, line 59: So we could record durations into a histogram intended for values; this seems silly for our ecosystem in which we value type safety for such stuff.

Sources/Prometheus/Docs.docc/swift-metrics.md
```diff
@@ -111,28 +111,19 @@ generated in a third party library.
...
 
-factory.defaultDurationHistogramBuckets = [
-  .milliseconds(5),
-  .milliseconds(10),
-  .milliseconds(25),
-  .milliseconds(50),
-  .milliseconds(100),
+factory.defaultTimerHistogramBuckets = [
+  0.005, 0.01, 0.025, 0.05, 0.1
 ]
```
ktoso, line 123: This feels like a step backwards as well tbh

fabianfett:
@ktoso I should have stressed in my pr description more that I don't love this either and mainly want to drive a discussion here.

Sources/Prometheus/Docs.docc/swift-metrics.md
FranzBusch: I agree that this feels like a step backwards

FranzBusch (commented):
Sorry, I was a bit hasty with my previous approval and I mostly looked at the change for the metrics factory. I do agree that for the plain prometheus APIs this is a step backwards.
What pain did it cause that we want to drop the generic?

fabianfett:
@ktoso @FranzBusch @tomerd

> What pain did it cause that we want to drop the generic?

Mostly requiring adopters to write generic code.

> This feels like a step backwards as well tbh

I agree here as well. However, if we decide that Histogram should be generic, we should probably make `Counter` and `Gauge` generic as well. wdyt?

FranzBusch:
> I agree here as well. However, if we decide that Histogram should be generic, we should probably make `Counter` and `Gauge` generic as well. wdyt?

Not sure. When would we use a `Duration` based `Counter` or `Gauge`? Maybe that's useful. From my understanding on this PR I thought we only had the generic abstractions to make the interop with Prometheus easier but not because of `swift-metrics`.

fabianfett:
> Not sure. When would we use a `Duration` based `Counter` or `Gauge`? Maybe that's useful. From my understanding on this PR I thought we only had the generic abstractions to make the interop with Prometheus easier but not because of `swift-metrics`.

For `Counter` and `Gauge` generics would be more about Int64 vs UInt64 vs Double.

fabianfett:
@ktoso 

> Metric values in OpenMetrics MUST be either floating points or integers. Note that ingestors of the format MAY only support float64. The non-real values NaN, +Inf and -Inf MUST be supported. NaN MUST NOT be considered a missing value, but it MAY be used to signal a division by zero.

https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#values

ktoso:
Yeah though that's about storage... so, maybe meet in the happy middle and keep the Duration histogram API but store in terms of Doubles if that's what you're after to simplify the implementation @fabianfett ?

ktoso:
I'm going to close this because it's been dormant for 2 years already -- we can always reopen if necessary

PR #110, opened 2024-02-13
Unregister the metric fully when last labelled metric is unregistered

simonjbeaumont:
## Motivation

The registry is designed to return the same metric instance for a previously registered metric with the same name if, and only if:
- the metric type matches;
- the metric label _keys_ match (it's fine for the labels to differ); and
- in the case of histograms, the buckets match.

By design, it throws a fatal error if these conditions do not hold for the metric name.

However, the registry also provides methods for unregistering metrics, such that they will not be emitted. A consequence of this unregistration is it allows for the metric name to be reused:

```swift
registry.unregisterCounter(registry.makeCounter(name: "name"))
_ = registry.makeGauge(name: "name") // does not crash
```

But the API has some unexpected behaviour when unregistering metrics that were registered with labels:

```swift
registry.unregisterCounter(registry.makeCounter(name: "name", labels: [("a", "1")]))
_ = registry.makeCounter(name: "name", labels: [("b", "1")])  // crashes
```

This is because the unregister APIs currently only remove the dimensions but do not go as far as to unregister the whole metric once the last such dimension has been unregistered.

## Modifications

- Unregister the metric fully when last labelled metric is unregistered
- Add test, `testUnregisterReregisterWithoutLabels`, which passed before this patch.
- Add test, `testUnregisterReregisterWithLabels`, which failed before this patch, but now passes.

## Result

Unregistering a metric with a label will unregister the metric name if this was the only remaining dimension for that metric.

### DIFF
--- a/Sources/Prometheus/PrometheusCollectorRegistry.swift
+++ b/Sources/Prometheus/PrometheusCollectorRegistry.swift
@@ -409,7 +409,11 @@ public final class PrometheusCollectorRegistry: Sendable {
                 let labelsKey = LabelsKey(counter.labels)
                 guard dimensions[labelsKey] === counter else { return }
                 dimensions.removeValue(forKey: labelsKey)
-                store[counter.name] = .counterWithLabels(labelNames, dimensions)
+                if dimensions.isEmpty {
+                    store.removeValue(forKey: counter.name)
+                } else {
+                    store[counter.name] = .counterWithLabels(labelNames, dimensions)
+                }
             default:
                 return
             }
@@ -431,7 +435,11 @@ public final class PrometheusCollectorRegistry: Sendable {
                 let dimensionsKey = LabelsKey(gauge.labels)
                 guard dimensions[dimensionsKey] === gauge else { return }
                 dimensions.removeValue(forKey: dimensionsKey)
-                store[gauge.name] = .gaugeWithLabels(labelNames, dimensions)
+                if dimensions.isEmpty {
+                    store.removeValue(forKey: gauge.name)
+                } else {
+                    store[gauge.name] = .gaugeWithLabels(labelNames, dimensions)
+                }
             default:
                 return
             }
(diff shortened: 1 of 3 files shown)


Sources/Prometheus/PrometheusCollectorRegistry.swift
```diff
@@ -409,6 +409,10 @@ public final class PrometheusCollectorRegistry: Sendable {
                 let labelsKey = LabelsKey(counter.labels)
                 guard dimensions[labelsKey] === counter else { return }
                 dimensions.removeValue(forKey: labelsKey)
+                guard !dimensions.isEmpty else {
+                    store.removeValue(forKey: counter.name)
+                    return
+                }
                 store[counter.name] = .counterWithLabels(labelNames, dimensions)
```
fabianfett, line 416: Personal preference nit, can we write his as:
```suggestion
                if dimensions.isEmpty {
                    store.removeValue(forKey: counter.name)
                } else {
                    store[counter.name] = .counterWithLabels(labelNames, dimensions)
                }
```
This way we remove a double negation, which should make this easier to reason about.

simonjbeaumont: I followed the double-negation guard pattern from the rest of this file (used in all the make APIs), e.g. 

```swift 
        guard !labels.isEmpty else {
            return self.makeValueHistogram(name: name, buckets: buckets)
        }
```

But I don't love it, so I'm happy to switch to an `if-else`.

fabianfett (approved):
Great fix! Thank you!

fabianfett:
@swift-server-bot test this please

fabianfett:
@swift-server-bot test this please

fabianfett:
@swift-server-bot test this please

fabianfett:
@swift-server-bot test this please

ktoso (commented):
Thanks, LGTM

ktoso (approved):

PR #130, opened 2025-07-31
new: add optional HELP line support

incertum:
Add optional HELP line support (includes help text validation and sanitization).

Follow for https://github.com/swift-server/swift-prometheus/pull/126

CC @ktoso @FranzBusch thanks in advance for your review!

### DIFF
--- a/Sources/Prometheus/MetricDescriptor.swift
+++ b/Sources/Prometheus/MetricDescriptor.swift
@@ -30,7 +30,7 @@ public struct MetricNameDescriptor {
     /// An optional suffix describing the metric's unit (e.g., `total`).
     public let unitName: String?
 
-    /// Optional descriptive text for the metric.
+    /// Optional help text for the metric. If a non-empty string is provided, it will be emitted as a `# HELP` line in the exposition format. If the parameter is omitted or an empty string is passed, the `# HELP` line will not be generated for this metric.
     public let helpText: String?
 
     /// Creates a new ``MetricNameDescriptor`` that defines the components of a fully qualified Prometheus metric name.
@@ -39,7 +39,7 @@ public struct MetricNameDescriptor {
     /// - Parameter subsystem: An optional subsystem to group related metrics within a namespace.
     /// - Parameter metricName: The required, descriptive base name of the metric.
     /// - Parameter unitName: An optional suffix describing the metric's unit (e.g., `total`).
-    /// - Parameter helpText: Optional descriptive text for the metric.
+    /// - Parameter helpText: Optional help text for the metric. If a non-empty string is provided, it will be emitted as a `# HELP` line in the exposition format. If the parameter is omitted or an empty string is passed, the `# HELP` line will not be generated for this metric.
     public init(
         namespace: String? = nil,
         subsystem: String? = nil,
@@ -55,7 +55,7 @@ public struct MetricNameDescriptor {
         self.helpText = helpText
     }
 
-    /// The fully qualified metric name, joining non-empty components with underscores.
+    /// The fully qualified metric name, joining non-empty components with underscores (e.g. `namespace_subsytem_metricName_unitName`).
     public var name: String {
         [namespace, subsystem, metricName, unitName]
             .compactMap { $0?.isEmpty == false ? $0 : nil }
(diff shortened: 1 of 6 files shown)


Sources/Prometheus/PrometheusCollectorRegistry.swift
```diff
@@ -71,16 +71,18 @@ public final class PrometheusCollectorRegistry: Sendable {
     /// created ``Counter`` will be part of the export.
     ///
     /// - Parameter name: A name to identify ``Counter``'s value.
+    /// - Parameter help: Help text for the metric. If a non-empty string is provided, it will be emitted as a `# HELP` line in the exposition format. If the parameter is omitted or an empty string is passed, the `# HELP` line will not be generated for this metric.
     /// - Returns: A ``Counter`` that is registered with this ``PrometheusCollectorRegistry``
-    public func makeCounter(name: String) -> Counter {
+    public func makeCounter(name: String, help: String = "") -> Counter {
```
incertum, line 76: Curious what you think would be the best approach to support the `HELP` line throughout without breaking things. Thought that a default value could solve it.

Sources/Prometheus/PrometheusCollectorRegistry.swift
```diff
@@ -780,6 +828,22 @@ extension String {
         return true
     }
 
+    fileprivate func isValidHelpText() -> Bool {
```
incertum, line 831: As a security person string sanitization always scares me as you usually get it wrong. Would you have better ideas here?

Sources/Prometheus/PrometheusCollectorRegistry.swift
ktoso: So _technically_ this isn't _100%_ source compatible because folks can get a function handle to a method and that then changes the signature of it which could break folks.

In practice, even Swift is breaking such APIs so given how not absolutely everywhere this API is used I think this is good enough IMHO -- it will be source compatible for the normal usage of calling methods. 

This package does not need to worry about ABI (which this does break), so that's not a problem.

incertum: Thanks for adding some context @ktoso -- super helpful! Should you come up with a better idea later, happy to quickly change it, just thought the HELP line option should also be available when not using the newer `MetricNameDescriptor`.

Sources/Prometheus/MetricDescriptor.swift
```diff
@@ -30,7 +30,7 @@ public struct MetricNameDescriptor {
     /// An optional suffix describing the metric's unit (e.g., `total`).
     public let unitName: String?
 
-    /// Optional descriptive text for the metric.
+    /// Optional help text for the metric. If a non-empty string is provided, it will be emitted as a `# HELP` line in the exposition format. If the parameter is omitted or an empty string is passed, the `# HELP` line will not be generated for this metric.
```
ktoso, line 33: That's a very long line, can we add some linebreaks please?

Sources/Prometheus/PrometheusCollectorRegistry.swift
ktoso: I guess we could `"....".allSatisfy { character in character.isLetter || character.isNumber }` but also allow punctuation etc, or lean into CharacterSet from Foundation...

ktoso (approved):
looks good, minor nitpick

Sources/Prometheus/PrometheusCollectorRegistry.swift
incertum: ACK, I'll check tmrw and possibly re-push the last commit and add that in.

Sources/Prometheus/MetricDescriptor.swift
incertum: The formatting let it pass though, yup will do!

Sources/Prometheus/PrometheusCollectorRegistry.swift
incertum: @ktoso besides adding all overloads, I refactored the `HelpText` sanitization a bit, however still without Foundation in order to limit some side-effects (such as increasing the binary size for adopters). 
Would you be able to review these changes? Much appreciated ❤️

ktoso (approved):
LGTM, thank you!

PR #131, opened 2025-08-05
wip: refactor: uniquely identify metric time series by name and label sets

incertum:
Previously, metrics were identified by their name. This change allows a shared metric name per metric type, with each unique set of labels defining a distinct time series.

Additionally, the internal data structure for a stored metric has been refactored, providing a more robust and programmatic representation.

CC @FranzBusch, @ktoso, @blindspotbounty

Let me know how you'd prefer to proceed. We have two main options:

1.  I can split this work into a few smaller, more focused PRs.
2.  We can finalize the overall approach in this PR, and I can push the remaining changes here next week.

Fixes https://github.com/swift-server/swift-prometheus/issues/125 opened by @blindspotbounty

### DIFF
--- a/Sources/Prometheus/PrometheusCollectorRegistry.swift
+++ b/Sources/Prometheus/PrometheusCollectorRegistry.swift
@@ -47,9 +47,14 @@ public final class PrometheusCollectorRegistry: Sendable {
         }
     }
 
-    private enum Metric {
-        case counter(Counter, help: String)
-        case counterWithLabels([String], [LabelsKey: Counter], help: String)
+    private typealias Metric<T> = (metric: T, help: String)
+
+    private struct MetricContainer<T> {
+        var metrics: [LabelsKey: Metric<T>]
+    }
+
+    private enum MetricType {
+        case counterMetrics(String, MetricContainer<Counter>)
         case gauge(Gauge, help: String)
         case gaugeWithLabels([String], [LabelsKey: Gauge], help: String)
         case durationHistogram(DurationHistogram, help: String)
@@ -58,7 +63,7 @@ public final class PrometheusCollectorRegistry: Sendable {
         case valueHistogramWithLabels([String], [LabelsKey: ValueHistogram], [Double], help: String)
     }
 
-    private let box = NIOLockedValueBox([String: Metric]())
+    private let box = NIOLockedValueBox([String: MetricType]())
 
     /// Create a new collector registry
     public init() {}
@@ -75,25 +80,7 @@ public final class PrometheusCollectorRegistry: Sendable {
     ///                   If the parameter is omitted or an empty string is passed, the `# HELP` line will not be generated for this metric.
     /// - Returns: A ``Counter`` that is registered with this ``PrometheusCollectorRegistry``
     public func makeCounter(name: String, help: String) -> Counter {
-        let name = name.ensureValidMetricName()
-        let help = help.ensureValidHelpText()
-        return self.box.withLockedValue { store -> Counter in
-            guard let value = store[name] else {
-                let counter = Counter(name: name, labels: [])
-                store[name] = .counter(counter, help: help)
-                return counter
-            }
-            guard case .counter(let counter, _) = value else {
-                fatalError(
-                    """
-                    Could not make Counter with name: \(name), since another metric type
-                    already exists for the same name.
-                    """
-                )
-            }
-
-            return counter
-        }
+        return self.makeCounter(name: name, labels: [], help: help)
     }
 
     /// Creates a new ``Counter`` collector or returns the already existing one with the same name.
@@ -104,7 +91,7 @@ public final class PrometheusCollectorRegistry: Sendable {
     /// - Parameter name: A name to identify ``Counter``'s value.
     /// - Returns: A ``Counter`` that is registered with this ``PrometheusCollectorRegistry``
     public func makeCounter(name: String) -> Counter {
-        return self.makeCounter(name: name, help: "")
+        return self.makeCounter(name: name, labels: [], help: "")
     }
 
     /// Creates a new ``Counter`` collector or returns the already existing one with the same name,
@@ -116,7 +103,7 @@ public final class PrometheusCollectorRegistry: Sendable {
     /// - Parameter descriptor: An ``MetricNameDescriptor`` that provides the fully qualified name for the metric.
     /// - Returns: A ``Counter`` that is registered with this ``PrometheusCollectorRegistry``
     public func makeCounter(descriptor: MetricNameDescriptor) -> Counter {
-        return self.makeCounter(name: descriptor.name, help: descriptor.helpText ?? "")
+        return self.makeCounter(name: descriptor.name, labels: [], help: descriptor.helpText ?? "")
     }
 
     /// Creates a new ``Counter`` collector or returns the already existing one with the same name.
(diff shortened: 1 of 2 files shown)


Sources/Prometheus/PrometheusCollectorRegistry.swift
```diff
@@ -47,9 +47,14 @@ public final class PrometheusCollectorRegistry: Sendable {
...
-        case counter(Counter, help: String)
-        case counterWithLabels([String], [LabelsKey: Counter], help: String)
+    private typealias Metric<T> = (metric: T, help: String)
+
+    private struct MetricContainer<T> {
+        var metrics: [LabelsKey: Metric<T>]
+    }
+
+    private enum MetricType {
+        case counterMetrics(String, MetricContainer<Counter>)
```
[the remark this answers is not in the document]
incertum: Also I would love to benefit from your Swift expertise re below ... it's probably the Type? It should be typed out for each metric type?

```
 note: consider making enum 'MetricType' conform to the 'Sendable' protocol
    private enum MetricType {
```

Sources/Prometheus/PrometheusCollectorRegistry.swift
```diff
@@ -47,9 +47,22 @@ public final class PrometheusCollectorRegistry: Sendable {
...
+        // most recent one. For an unlabelled Counter, the empty label set is used as the key, and the
+        // collection contains only one entry. Finally, for clarification, the same Counter metric name can
+        // simultaneously be labeled and unlabeled.
+        var countersByLabelSets: [LabelsKey: CounterWithHelp]
+    }
+
     private enum Metric {
-        case counter(Counter, help: String)
-        case counterWithLabels([String], [LabelsKey: Counter], help: String)
+        case counter(name: String, CounterGroup)
```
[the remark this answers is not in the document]
incertum: It seemed that way before for labeled cases, so I thought it was a design preference. 

Based on your comment, I understand that you approve of removing the small duplicative part. I’ve just re-pushed it.

Could you let me know if the design is now ok? If so, I’ll proceed to implement the changes for Gauge in this PR as well.

Sources/Prometheus/PrometheusCollectorRegistry.swift
```diff
@@ -47,11 +47,33 @@ public final class PrometheusCollectorRegistry: Sendable {
         }
     }
 
+    // Note: In order to support Sendable, need to explicitely define the types.
```
ktoso, line 50: slightly weird note; we're not marking those Sendable at all? Remove the note?
```suggestion
```

incertum: Was intended to avoid someone proposing something like I had at the beginning, see https://github.com/swift-server/swift-prometheus/pull/131#discussion_r2252882330. Below would not work here:

```
private struct MetricGroup<T> {
    var metricsByLabelSets: [LabelsKey: T]
}
```

It may not be a good note; still remove or adjust it?

ktoso:
This could use some more use of generics rather than copying types around 1:1; here's a patch:
[0001-make-metric-groups-use-more-generics.patch](https://github.com/user-attachments/files/21978461/0001-make-metric-groups-use-more-generics.patch)

Basically:

```diff
-    // Note: In order to support Sendable, need to explicitely define the types.
-    private struct CounterWithHelp {
-        var counter: Counter
+    private struct MetricWithHelp<Metric: AnyObject & Sendable>: Sendable {
+        var metric: Metric
         let help: String
     }

-    private struct GaugeWithHelp {
-        var gauge: Gauge
-        let help: String
-    }
-
-    private struct CounterGroup {
-        // A collection of Counter metrics, each with a unique label set, that share the same metric name.
-        // Distinct help strings for the same metric name are permitted, but Prometheus retains only the
-        // most recent one. For an unlabelled Counter, the empty label set is used as the key, and the
-        // collection contains only one entry. Finally, for clarification, the same Counter metric name can
-        // simultaneously be labeled and unlabeled.
-        var countersByLabelSets: [LabelsKey: CounterWithHelp]
-    }
-
-    private struct GaugeGroup {
-        var gaugesByLabelSets: [LabelsKey: GaugeWithHelp]
+    /// A collection of metrics, each with a unique label set, that share the same metric name.
+    /// Distinct help strings for the same metric name are permitted, but Prometheus retains only the
+    /// most recent one. For an unlabelled metric, the empty label set is used as the key, and the
+    /// collection contains only one entry. Finally, for clarification, the same metric name can
+    /// simultaneously be labeled and unlabeled.
+    private struct MetricGroup<Metric: Sendable & AnyObject>: Sendable {
+        var metricsByLabelSets: [LabelsKey: MetricWithHelp<Metric>]
     }

     private enum Metric {
-        case counter(CounterGroup)
-        case gauge(GaugeGroup)
+        case counter(MetricGroup<Counter>)
+        case gauge(MetricGroup<Gauge>)
```

Sources/Prometheus/PrometheusCollectorRegistry.swift
ktoso: Yeah, something like that. I proposed a patch below; we can converge those types a bit more and annotate them as Sendable as well: https://github.com/swift-server/swift-prometheus/pull/131#issuecomment-3222165204

incertum: Ah, I see where I was wrong with my initial attempt. Awesome, glad this suboptimal note got us to re-review this part!

ktoso (approved):
Looks good to me, thank you.

incertum:
> Have you tested that this is correctly imported by a prometheus importer? IIRC, prometheus importer doesn't allow flexible labels per metric name. IIRC this means for a given metric name the label names must always be the same.

Based on the new documentation example provided:

```sh
# HELP myapp_http_connections_active Currently active HTTP connections
# TYPE myapp_http_connections_active gauge
myapp_http_connections_active 42.0
# HELP myapp_http_requests_total Total HTTP requests
# TYPE myapp_http_requests_total counter
myapp_http_requests_total{method="GET",status="200"} 5.0
myapp_http_requests_total{method="POST",status="201"} 2.0
myapp_http_requests_total{method="GET",status="500",endpoint="/api/users"} 1
# HELP myapp_system_errors_total Total system errors
# TYPE myapp_system_errors_total counter
myapp_system_errors_total 3.0
# HELP myapp_http_request_duration_seconds HTTP request duration in seconds
# TYPE myapp_http_request_duration_seconds histogram
myapp_http_request_duration_seconds_bucket{le="0.01"} 0
myapp_http_request_duration_seconds_bucket{le="0.1"} 0
myapp_http_request_duration_seconds_bucket{le="1.0"} 1
myapp_http_request_duration_seconds_bucket{le="+Inf"} 1
myapp_http_request_duration_seconds_sum 0.15
myapp_http_request_duration_seconds_count 1
# HELP myapp_system_memory_usage_bytes Current memory usage in bytes
# TYPE myapp_system_memory_usage_bytes gauge
myapp_system_memory_usage_bytes 134217728.0
```

- Scrapping confirmed to be successful, with no errors.
- Prometheus aggregation confirmed to be correct. For instance, `sum(myapp_http_requests_total)` yields the correct result of 8, and `sum by (method) (myapp_http_requests_total)` also yields the correct aggregation.

> ### Update
> Found the statement in the prometheus docs:
> 
> > Client libraries MUST NOT allow users to have different label names for the same metric for Gauge/Counter/Summary/Histogram or any other Collector offered by the library.
> 
> https://prometheus.io/docs/instrumenting/writing_clientlibs/#labels
> 
> Because of this I think we should revert #131 and #134.

The changes addressed the following requests from adopters: https://github.com/swift-server/swift-prometheus/issues/125 as follow to a java client discussion https://github.com/prometheus/client_java/issues/696

There may be some room for flexibility to support adopter use cases while avoiding common mistakes (e.g., enforcing that myapp_http_requests_total can’t be used unlabeled simultaneously to avoid incorrect aggregations as the sum should be computed in Prometheus).

For instance, also consider the following issue:
https://github.com/swift-server/swift-prometheus/issues/137
We currently don’t require a HELP text because doing so would break backwards compatibility. In fact, it wasn’t even supported before the 2.20 release. I’ve also noticed that the HELP line is declared as optional in some documentation places while it’s declared as MUST in others.

Perhaps we can get a definitive answer as we approach the Prometheus community. Let’s come back to this once we have a list of required refactors, if applicable. CC @ktoso.

PR #106, opened 2023-10-13
Add Benchmarks for metrics

freef4ll:
# Motivation

Adding benchmark tests in order to aid development and avoid performance regressions.

Benchmark is added in the same fashion as for  [swift-nio](https://github.com/apple/swift-nio/pull/2536) and [swift-certifcates](https://github.com/apple/swift-certificates/pull/125).

# Modifications

Benchmarks are added for incrementing a counter, gauge,  histogram and an export of 5000 metrics from the esystem.

# How to execute benchmark
```
cd Benchmarks
swift package benchmark
```
In Xcode can open the project like and use Instruments to profile:
```
open --env BENCHMARK_DISABLE_JEMALLOC=true Package.swift
```

# Result

Benchmarks are available, and performance difference between v1 is great which only managed to increment 2-3k times per second.

PR #133, opened 2025-08-26
Push 'Add Benchmarks for metrics #106' over the finish line

ktoso:
Attempt to take _Add Benchmarks for metrics_ #106 over the finish line; just some minor things were pending


ktoso:
Good point, we can bump minimum to 5.10 indeed, that's the "three versions back" we support

PR #134, opened 2025-08-26
refactor[Histograms]: uniquely identify metric time series by name and label sets (2/2)

incertum:
refactor: uniquely identify {Value,Duration}Histograms by name and label sets
    
Previously, {Value,Duration}Histograms could only be identified by their name. This change allows multiple {Value,Duration}Histograms to share the same name, with each unique set of labels defining a distinct time series.

However, the buckets are immutable for a MetricGroup once initialized with the first metric.

Additionally, the internal data structure for a stored metric has been refactored, providing a more robust and programmatic representation.

CC @ktoso @FranzBusch 

Follow for https://github.com/swift-server/swift-prometheus/pull/131, fixes https://github.com/swift-server/swift-prometheus/issues/125

### DIFF
--- a/Sources/Prometheus/PrometheusCollectorRegistry.swift
+++ b/Sources/Prometheus/PrometheusCollectorRegistry.swift
@@ -52,22 +52,38 @@ public final class PrometheusCollectorRegistry: Sendable {
         let help: String
     }
 
+    private enum BucketType: Sendable, Hashable {
+        case duration([Duration])
+        case value([Double])
+    }
+
+    /// A collection of metrics, each with a unique label set, that share the same metric name.
+    /// Distinct help strings for the same metric name are permitted, but Prometheus retains only the
+    /// first one. For an unlabelled metric, the empty label set is used as the key, and the
+    /// collection contains only one entry. Finally, for clarification, the same metric name can
+    /// simultaneously be labeled and unlabeled.
+    /// For histograms, the buckets are immutable for a MetricGroup once initialized with the first
+    /// metric. See also https://github.com/prometheus/OpenMetrics/issues/197.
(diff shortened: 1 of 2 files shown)


Sources/Prometheus/PrometheusCollectorRegistry.swift
```diff
@@ -52,22 +52,38 @@ public final class PrometheusCollectorRegistry: Sendable {
         let help: String
     }
 
+    private enum BucketType: Sendable, Hashable {
```
ktoso, line 55: Doesn't feel like a "type" it is the actual buckets - plural because of collection of bucket which is one value in here

```suggestion
    private enum HistogramBuckets: Sendable, Hashable {
```

ktoso (approved):
Looks good :)

Tests/PrometheusTests/HistogramTests.swift
```diff
@@ -371,6 +371,288 @@ final class HistogramTests: XCTestCase {
...
+            help: "Base metric name with a different label set variant"
+        )
+
+        var buffer = [UInt8]()
+        histogram0.recordNanoseconds(300_000_000)  // 300ms
+        histogram1.recordNanoseconds(600_000_000)  // 600ms
+        histogram2.recordNanoseconds(150_000_000)  // 150ms
+        histogram1.recordNanoseconds(1_500_000_000)  // 1500ms
+        histogram0.recordNanoseconds(800_000_000)  // 800ms
+        histogram2.recordNanoseconds(100_000_000)  // 100ms
```
ktoso, line 411: This is easier to write as
```
histogram.record(.milliseconds(100))
```

PR #135, opened 2025-08-27
feat: add {emitToString, emitToBuffer} options + dedup TYPE and HELP per metric name

incertum:
Add {emitToString, emitToBuffer} thread-safe internal buffer management variants to emit metrics.

Deduplicate TYPE and HELP lines per metric name by default on emit.

Fixes https://github.com/swift-server/swift-prometheus/issues/118 @ptoffy

CC @ktoso and @FranzBusch

### DIFF
--- a/Sources/Prometheus/PrometheusCollectorRegistry.swift
+++ b/Sources/Prometheus/PrometheusCollectorRegistry.swift
@@ -60,10 +60,18 @@ public final class PrometheusCollectorRegistry: Sendable {
     /// For histograms, the buckets are immutable for a MetricFamily once initialized with the first
     /// metric. See also https://github.com/prometheus/OpenMetrics/issues/197.
     private struct MetricFamily<Metric: Sendable & AnyObject>: Sendable {
-        var metricsByLabelSets: [LabelsKey: Metric]
+        private enum State {
+            case labeled([LabelsKey: Metric])
+            case unlabeled(Metric)
+            case empty
+        }
+
+        let metricsByLabelSets: [LabelsKey: Metric]
         let buckets: HistogramBuckets?
         let help: String?
         let metricUnlabeled: Metric?
+
+        private let state: State
 
         init(
             metricsByLabelSets: [LabelsKey: Metric] = [:],
@@ -71,10 +79,53 @@ public final class PrometheusCollectorRegistry: Sendable {
             help: String? = nil,
             metricUnlabeled: Metric? = nil
         ) {
+            // Validate mutual exclusivity on creation.
+            if metricUnlabeled != nil && !metricsByLabelSets.isEmpty {
+                fatalError("Cannot have both labeled and unlabeled metrics in the same family.")
+            }
+
             self.metricsByLabelSets = metricsByLabelSets
             self.buckets = buckets
             self.help = help
             self.metricUnlabeled = metricUnlabeled
+
+            // Set internal state for validation.
+            if let unlabeled = metricUnlabeled {
+                self.state = .unlabeled(unlabeled)
+            } else if !metricsByLabelSets.isEmpty {
+                self.state = .labeled(metricsByLabelSets)
+            } else {
+                self.state = .empty
+            }
+        }
+
+        func add(metric: Metric, for labels: [(String, String)]) -> MetricFamily<Metric> {
+            // This method should only be called for labeled metrics.
+            guard !labels.isEmpty else {
+                fatalError("Use initializer for unlabeled metrics, not add method")
+            }
+
+            switch state {
+            case .unlabeled:
+                fatalError("Cannot register a labeled metric when an unlabeled metric already exists.")
+            case .labeled, .empty:
+                break
+            }
+
+            let labelsKey = LabelsKey(labels)
+            guard metricsByLabelSets[labelsKey] == nil else {
+                return self
+            }
+
+            var newMetricsByLabelSets = metricsByLabelSets
+            newMetricsByLabelSets[labelsKey] = metric
+
+            return MetricFamily(
+                metricsByLabelSets: newMetricsByLabelSets,
+                buckets: buckets,
+                help: help,
+                metricUnlabeled: metricUnlabeled
+            )
         }
     }
 
@@ -162,27 +213,7 @@ public final class PrometheusCollectorRegistry: Sendable {
             }
 
             switch entry {
-            case .counter(var existingCounterFamily):
-                // Check mutual exclusivity constraints. While Prometheus wouldn't break with sharing a
-                // metric name for labeled and unlabeled variants, the client enforces that each unique
-                // metric name must either be labeled (allowing multiple label sets) or unlabeled
-                // in order to support correct Prometheus aggregations.
-                if existingCounterFamily.metricUnlabeled != nil && !labels.isEmpty {
-                    fatalError(
-                        """
-                        Label conflict for metric '\(name)':
-                        Cannot register a labeled metric when an unlabeled metric already exists.
-                        """
-                    )
-                }
-                if labels.isEmpty && !existingCounterFamily.metricsByLabelSets.isEmpty {
-                    fatalError(
-                        """
-                        Label conflict for metric '\(name)':
-                        Cannot register an unlabeled metric when labeled metrics already exist.
-                        """
-                    )
-                }
+            case .counter(let existingCounterFamily):
 
                 // For unlabeled metrics, return the existing one if it exists.
                 if labels.isEmpty, let existingUnlabeled = existingCounterFamily.metricUnlabeled {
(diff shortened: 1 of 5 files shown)


Sources/Prometheus/PrometheusCollectorRegistry.swift
```diff
@@ -82,9 +82,25 @@ public final class PrometheusCollectorRegistry: Sendable {
...
     private let box = NIOLockedValueBox([String: Metric]())
+    private let dedupTypeHelpPerMetricNameOnEmit: Bool
 
-    /// Create a new collector registry
-    public init() {}
+    /// Creates a new PrometheusCollectorRegistry with default configuration.
+    ///
+    /// Uses deduplication for TYPE and HELP lines according to Prometheus specifications,
+    /// where only one TYPE and HELP line is emitted per metric name regardless of label sets.
+    public init() {
```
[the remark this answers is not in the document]
incertum: Additional note: Since the last release, we have not allowed reusing the same metric name for distinct label sets, hence this new default should not break anything.

Sources/Prometheus/PrometheusCollectorRegistry.swift
```diff
@@ -82,9 +82,25 @@ public final class PrometheusCollectorRegistry: Sendable {
...
+        self.dedupTypeHelpPerMetricNameOnEmit = true
```
incertum, line 92: Additional note: Since the last release, we have not allowed reusing the same metric name for distinct label sets, hence this new default should not break anything.

Sources/Prometheus/PrometheusCollectorRegistry.swift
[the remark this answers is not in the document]
ktoso: To clarify, this package has source compat but no ABI compat, but yes, we need it still :)

incertum:
Fixes https://github.com/swift-server/swift-prometheus/issues/137

incertum:
Fixes https://github.com/swift-server/swift-prometheus/issues/138, amendment to https://github.com/swift-server/swift-prometheus/issues/125 as initial fixes were slightly too permissive wrt metric name sharing.

Sources/Prometheus/PrometheusCollectorRegistry.swift
```diff
@@ -145,26 +152,66 @@ public final class PrometheusCollectorRegistry: Sendable {
...
             switch entry {
-            case .counter(var existingCounterGroup):
-                if let existingCounterWithHelp = existingCounterGroup.metricsByLabelSets[key] {
-                    return existingCounterWithHelp.metric
+            case .counter(var existingCounterFamily):
+                // Check mutual exclusivity constraints. While Prometheus wouldn't break with sharing a
+                // metric name for labeled and unlabeled variants, the client enforces that each unique
+                // metric name must either be labeled (allowing multiple label sets) or unlabeled
+                // in order to support correct Prometheus aggregations.
+                if existingCounterFamily.metricUnlabeled != nil && !labels.isEmpty {
```
ktoso, line 170: Can't we do this in the constructor of `MetricFamily`? otherwise there's still ways to construct an illegal Family by accident

ktoso: I'd probably make all propertires `let` in the family so we can't violate the rule by external mutation, and then make some method called amend or add that performs these checks inside there. This way we don't have a chance to get the type into illegal state

Sources/Prometheus/PrometheusCollectorRegistry.swift
```diff
@@ -266,26 +313,66 @@ public final class PrometheusCollectorRegistry: Sendable {
...
             switch entry {
-            case .gauge(var existingGaugeGroup):
-                if let existingGaugeWithHelp = existingGaugeGroup.metricsByLabelSets[key] {
-                    return existingGaugeWithHelp.metric
+            case .gauge(var existingGaugeFamily):
+                // Check mutual exclusivity constraints. While Prometheus wouldn't break with sharing a
+                // metric name for labeled and unlabeled variants, the client enforces that each unique
+                // metric name must either be labeled (allowing multiple label sets) or unlabeled
+                // in order to support correct Prometheus aggregations.
+                if existingGaugeFamily.metricUnlabeled != nil && !labels.isEmpty {
```
ktoso, line 331: yeah if we centralize this series of checks in the Family type we don't have to duplicate this here, wdyt

Sources/Prometheus/PrometheusCollectorRegistry.swift
```diff
@@ -790,59 +991,163 @@ public final class PrometheusCollectorRegistry: Sendable {
...
+    /// Uses an internal buffer that auto-sizes on first call to find optimal initial capacity. The buffer
+    /// may resize during the registry's lifetime if output grows significantly. Subsequent calls reuse the
+    /// established capacity, clearing content but preserving the initially allocated memory.
+    ///
+    /// - Returns: A String containing all registered metrics in Prometheus text format
+    /// - Note: Thread-safe. Use ``resetInternalBuffer()`` to force recalibration
+    /// - SeeAlso: ``emitToBuffer()`` for raw UTF-8 bytes, ``emit(into:)`` for custom buffer
+    public func emitToString() -> String {
+        return bufferBox.withLockedValue { buffer in
+            guard buffer.capacity == 0 else {
```
ktoso, line 1035: I'd rather have a plain if else here because this isn't really guard + early exit but _actual logic_ in the else case; which is better with an if/else

Sources/Prometheus/PrometheusCollectorRegistry.swift
```diff
@@ -47,43 +47,50 @@ public final class PrometheusCollectorRegistry: Sendable {
...
 
-    /// A collection of metrics, each with a unique label set, that share the same metric name.
-    /// Distinct help strings for the same metric name are permitted, but Prometheus retains only the
-    /// first one. For an unlabelled metric, the empty label set is used as the key, and the
-    /// collection contains only one entry. Finally, for clarification, the same metric name can
-    /// simultaneously be labeled and unlabeled.
-    /// For histograms, the buckets are immutable for a MetricGroup once initialized with the first
+    /// A MetricFamily.
+    ///
+    /// A metric name can either map to multiple labeled metrics OR a single unlabeled metric,
```
ktoso, line 57: Since this is a strict requirement we should make sure it can't be violated.

I'd make the metricsByLabelSets a `let`, or make it `private`, and make a new method through which we modify or make a new family instance; there we'd centralize all the checks.

You could also do a small enum for the two cases if you want inside of the MetricFamily `enum Metrics { case labelled([LabelsKey: Metric]); case unlabelled(Metric) }` which will help to avoid violating the rule 👀

incertum: True 😉 pushed a suggestion that hopefully addresses all comments in this regard!

Sources/Prometheus/PrometheusCollectorRegistry.swift
incertum: Turns out the formatting `curl -s --retry 3 https://raw.githubusercontent.com/swiftlang/github-workflows/refs/heads/main/.github/workflows/scripts/check-swift-format.sh | bash`...  reverts it again 🤷‍♀️

ktoso: It changes if to guard, wat! Our rules are overdoing it clearly then… but let’s not hold up the PR by this

Sources/Prometheus/PrometheusCollectorRegistry.swift
```diff
@@ -60,21 +60,72 @@ public final class PrometheusCollectorRegistry: Sendable {
...
+            if let unlabeled = metricUnlabeled {
+                self.state = .unlabeled(unlabeled)
+            } else if !metricsByLabelSets.isEmpty {
+                self.state = .labeled(metricsByLabelSets)
+            } else {
+                self.state = .empty
+            }
+        }
+
+        func add(metric: Metric, for labels: [(String, String)]) -> MetricFamily<Metric> {
```
ktoso, line 102: Minor and doesn’t have to be done, but fyi. In swift often the “add” would be mutating and the not mutating one would be “adding”. But this is fine, just internal api so no need to bikeshed too much

Sources/Prometheus/PrometheusCollectorRegistry.swift
```diff
@@ -60,21 +60,72 @@ public final class PrometheusCollectorRegistry: Sendable {
...
     /// metric. See also https://github.com/prometheus/OpenMetrics/issues/197.
     private struct MetricFamily<Metric: Sendable & AnyObject>: Sendable {
-        var metricsByLabelSets: [LabelsKey: Metric]
+        private enum State {
+            case labeled([LabelsKey: Metric])
+            case unlabeled(Metric)
+            case empty
+        }
+
+        let metricsByLabelSets: [LabelsKey: Metric]
```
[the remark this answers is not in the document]
incertum: Pushed another suggestion to further reduce code duplication throughout.

ktoso (approved):
Lgtm!

PR #140, opened 2025-09-03
Concurrency: enable 6.0 language mode / concurrency safety

ktoso:
Swift 6 language mode enables complete concurrency checking and will error when races are found.

Also fix a few types to be 6 language mode compatible.

### DIFF
--- a/Sources/Prometheus/NIOLock.swift
+++ b/Sources/Prometheus/NIOLock.swift
@@ -216,13 +216,13 @@ extension LockStorage: Sendable {}
 /// of lock is safe to use with `libpthread`-based threading models, such as the
 /// one used by NIO. On Windows, the lock is based on the substantially similar
 /// `SRWLOCK` type.
-public struct NIOLock {
+struct NIOLock {
     @usableFromInline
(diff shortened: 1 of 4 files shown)


Sources/Prometheus/NIOLock.swift
```diff
@@ -189,23 +201,28 @@ final class LockStorage<Value>: ManagedBuffer<Value, LockPrimitive> {
...
 ///
-/// - note: ``NIOLock`` has reference semantics.
+/// - Note: ``NIOLock`` has reference semantics.
 ///
 /// This object provides a lock on top of a single `pthread_mutex_t`. This kind
 /// of lock is safe to use with `libpthread`-based threading models, such as the
 /// one used by NIO. On Windows, the lock is based on the substantially similar
 /// `SRWLOCK` type.
-struct NIOLock {
+public struct NIOLock {
```
FranzBusch, line 219: I don't think that we should make this type nor the other methods public

ktoso: Yeah, overlooked that just copying the latest from bio — thanks.

PR #144, opened 2025-09-17
Update README with Swift Metrics backend info

ktoso:
Callout that this is swift metrics compatible in the readme


ktoso:
Just adding a swift metrics mention so we tag a release and start chatting with upstream maintainers :)

PR #147, opened 2025-09-26
Add `Package.resolved` and `dependabot.yml`

incertum:
Add `Package.resolved` and `dependabot.yml`

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+---
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+  - package-ecosystem: swift
+    directory: /
+    schedule:
+      interval: weekly
+  - package-ecosystem: swift
+    directory: /Benchmarks
+    schedule:
+      interval: weekly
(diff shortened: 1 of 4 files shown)


FranzBusch (changes requested):
I don't think that we should go ahead with this PR. The recommendation right now is to not commit the resolved file in libraries since the file has no impact on users of the package. Furthermore, libraries should avoid bumping the min versions unless there is a need for it e.g. new APIs available in a dependency.

incertum:
> I don't think that we should go ahead with this PR. The recommendation right now is to not commit the resolved file in libraries since the file has no impact on users of the package. Furthermore, libraries should avoid bumping the min versions unless there is a need for it e.g. new APIs available in a dependency.

There is a bigger discussion ongoing around this. Let's move this PR to draft state until there are new features or a new decision / outcome in this regard.

ktoso:
Yeah, it is not useful to have a Package.resolved in libraries.

Package.resolved isn't a real lock file; it doesn't matter at all in libraries either since consumer will just resolve anyway.

ktoso (changes requested):
I don't think we should do this in libraries

PR #151, opened 2025-09-30
update: set explicit permissions in GitHub workflows (1/2)

incertum:
update: set explicit permissions in GitHub workflows (1/2)

PR 1: pull_request related workflows only

See https://github.com/swift-server/swift-prometheus/issues/150

CC @FranzBusch


ktoso (approved):
lgtm

ktoso:
I guess that's a pattern we wanna do everywhere huh

PR #171, opened 2026-06-12
Add environment variable configuration for default histogram buckets

### LINKED ISSUE #142: Support a env var to configure default histogram buckets
It may be common for service operators to try to tune the histogram buckets by trying it out in practice.

It may be nice to allow configuring this by `SWIFT_PROMETHEUS_DEFAULT_HISTOGRAMS=0.1,0.2,0.5,.75,0.9` that would configure the bucket size unless they are configured explicitly in code.

armandgrillet:
Fixes #142.

## Summary

This adds environment-variable configuration for the default histogram buckets used by `PrometheusMetricsFactory`.

The main variable is:

- `SWIFT_PROMETHEUS_DEFAULT_HISTOGRAMS`

This matches the variable proposed in #142 and applies to both default duration histogram buckets and default value histogram buckets. Values are interpreted as seconds for duration histograms and as raw upper bounds for value histograms.

Because Swift Prometheus exposes separate defaults for duration and value histograms, this PR also adds more specific override variables:

- `SWIFT_PROMETHEUS_DEFAULT_DURATION_HISTOGRAM_BUCKETS`
- `SWIFT_PROMETHEUS_DEFAULT_VALUE_HISTOGRAM_BUCKETS`

Those typed variables take precedence over `SWIFT_PROMETHEUS_DEFAULT_HISTOGRAMS` when set to valid values.

Explicit configuration in code still takes precedence over environment-derived defaults.

## Validation

Malformed values are ignored as if the variable were unset: a malformed specific variable falls back to `SWIFT_PROMETHEUS_DEFAULT_HISTOGRAMS`, then to the built-in defaults. The parser rejects:

- empty lists
- non-finite values
- non-numeric values
- repeated or decreasing bucket upper bounds
- empty comma-separated elements such as `1,,2`
- negative duration bucket upper bounds
- duration values too large to represent as a `Duration` (converting them would trap, so environment input cannot crash the process)

## Testing

- `swift test`
- `git diff --check origin/main...HEAD`

### DIFF
--- a/Sources/Prometheus/Docs.docc/swift-metrics.md
+++ b/Sources/Prometheus/Docs.docc/swift-metrics.md
@@ -171,4 +171,31 @@ Now a `Timer` with the label "long" will use the buckets
 
 The same functionality is also available for ``ValueHistogram`` and aggregating `Recorder`s.
 
+Buckets from the environment:
+
+The default buckets can also be set with environment variables, allowing operators to tune the
+default histogram buckets without code changes:
+
+```bash
+export SWIFT_PROMETHEUS_DEFAULT_HISTOGRAMS=0.05,0.1,0.25,0.5,1,5
+```
+
+Values from `SWIFT_PROMETHEUS_DEFAULT_HISTOGRAMS` are interpreted as seconds for duration
+histograms and as raw upper bounds for value histograms. Bucket upper bounds must be finite,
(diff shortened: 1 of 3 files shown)


FranzBusch:
Maybe we should adopt `swift-configuration` instead here which will allow both env variable but also many other formats for configuration.

